PR #19, opened 2024-11-07
fix(win): set border to none for backdrop windows

1borgy:
## Description
Previously, backdrop windows were being created with the default border options set in `opts`. I can't imagine this is ever desirable (see screenshots), so always set the backdrop window border to none.

## Related Issue(s)
N/A

### Before:
image

### After:
image

### DIFF
--- a/lua/snacks/win.lua
+++ b/lua/snacks/win.lua
@@ -429,6 +429,7 @@ function M:drop()
       height = 0,
       width = 0,
       style = "minimal",
+      border = "none",
       focusable = false,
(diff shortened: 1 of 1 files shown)


folke:
Missed that and my colorscheme didn't really show a border. ty!

1borgy:
> Missed that and my colorscheme didn't really show a border. ty!

No problem! Having a lot of fun playing around with snacks - has a lot of my wishlist items from LazyVim! Truly appreciate the work you do.

PR #23, opened 2024-11-07
feat(words): `jump` optionally shows notification with reference count

chrisgrieser:
## Description

Add an opt-in option to show a notification for `Snacks.words.jump`, that displays the number and index of the current reference. The notification uses `Snacks.notify.info` to be consistent with the rest of the repo.

## Screenshots

Pasted image 2024-11-07 at 16 39 46


folke:
Made some changes and merged. Thanks!

PR #31, opened 2024-11-08
feat(words): add `fold_open` and `set_jump_point` config options

chrisgrieser:
## Description
Added `fold_open` and `set_jump_point` as config options for `Snacks.words.jump`.

### DIFF
--- a/docs/words.md
+++ b/docs/words.md
@@ -13,6 +13,8 @@ Auto-show LSP references and quickly navigate between them
   debounce = 200, -- time in ms to wait before updating
   notify_jump = false, -- show a notification when jumping
   notify_end = true, -- show a notification when reaching the end
+  fold_open = true, -- open folds after jumping
+  set_jump_point = true, -- set jump point before jumping
   modes = { "n", "i", "c" }, -- modes to show references
 }
 ```
--- a/docs/words.md
+++ b/docs/words.md
@@ -14,7 +14,7 @@ Auto-show LSP references and quickly navigate between them
   notify_jump = false, -- show a notification when jumping
   notify_end = true, -- show a notification when reaching the end
   fold_open = true, -- open folds after jumping
-  set_jump_point = true, -- set jump point before jumping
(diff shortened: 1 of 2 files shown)


docs/words.md
```diff
@@ -13,6 +13,8 @@ Auto-show LSP references and quickly navigate between them
   debounce = 200, -- time in ms to wait before updating
   notify_jump = false, -- show a notification when jumping
   notify_end = true, -- show a notification when reaching the end
+  fold_open = true, -- open folds after jumping
+  set_jump_point = true, -- set jump point before jumping
```
folke, line 17: Can you rename this to `jumplist`?

docs/words.md
```diff
@@ -13,6 +13,8 @@ Auto-show LSP references and quickly navigate between them
   debounce = 200, -- time in ms to wait before updating
   notify_jump = false, -- show a notification when jumping
   notify_end = true, -- show a notification when reaching the end
+  fold_open = true, -- open folds after jumping
```
folke, line 16: rename to `folds`?

folke (changes requested):
Great! Just two renames :)

chrisgrieser:
> Actually, maybe just remove the folds option and just always open folds.
I don't really see why someone would not have that option enabled...

Well, there is an option `:h foldopen`, so the behavior is configurable normally. Also, some people may prefer not to open all folds when they cycle the whole file

docs/words.md
chrisgrieser: sure about that? `fold_open` is quite verbose, while `folds` does not telegraph directly what the options does.

docs/words.md
chrisgrieser: sure

docs/words.md
folke: you're right. Maybe `foldopen` to resemble the option and cmd?

chrisgrieser: 👍

folke:
Awesome, thank you!

PR #47, opened 2024-11-09
docs(example): fix rename file keymap

jdujava:
## Description

Fixes "Rename File" keymap in Usage section of README.


folke:
ty

PR #64, opened 2024-11-12
fix(win): take border into account for window position

hareki:
## Description

The window position is not correctly centered when position is set to `float` and border isn't `none`. I found [a similar issue](https://github.com/folke/lazy.nvim/issues/812) in [lazy.nvim](https://github.com/folke/lazy.nvim) along with [your fix](https://github.com/folke/lazy.nvim/commit/451f217e9b2d71f08bdae0ce5ac7e8e8a6503f48) there, so I have applied the same fix here. Thank you for your work!

## Related Issue(s)
https://github.com/folke/lazy.nvim/issues/812

### Before
image

### After
image


folke:
ty

PR #66, opened 2024-11-12
fix(notifier): disable `colorcolumn` by default

chrisgrieser:
## Description
`colorcolumn` should be disabled by default, since in most cases, the user will not want to have them displayed in a notification.

## Screenshots
before:
Pasted image 2024-11-12 at 19 59 40

after:
Pasted image 2024-11-12 at 20 02 06


folke:
makes sense. Thanks!

PR #73, opened 2024-11-14
feat(terminal): added function `open_or_focus`

maomao9-0:
## Description

Add function to terminal.lua so that it has additional functionality to be able to focus on the terminal if it is already open; otherwise, open a new terminal.

### DIFF
--- a/lua/snacks/terminal.lua
+++ b/lua/snacks/terminal.lua
@@ -133,6 +133,27 @@ function M.toggle(cmd, opts)
   return terminals[id]
 end
 
+--- Focus terminal window if already open. Otherwise, open terminal window.
+--- The terminal id is based on the `cmd`, `cwd`, `env` and `vim.v.count1` options.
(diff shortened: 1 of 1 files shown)


maomao9-0:
The purpose of this feature is to allow people to make use of this function to run commands in terminal. For example, I am adding a keybind to allow me to compile my c++ code and run the executable file. However, if I use the default Snacks.terminal(), the default functionality is to toggle the terminal. So if the terminal is already open, my keybind will close the terminal and type the compilation commands into my c++ file instead of the terminal.
Having this `open_or_focus` function will allow my keybind to type the compilation commands into the terminal regardless of whether the terminal is already open or close.

folke:
You can already just do `:Snacks.terminal.toggle():show()` to achieve the same.

maomao9-0:
Oh I did not think of that. Thanks so much!

PR #78, opened 2024-11-15
fix(docs): typo in README.md

nikeedev:
Updated it because somebody place ! outside of the brackets :))

## Description

The big picture? Well, people can get lost because what does ![tip] mean compared to **<span style="background-color: green; border: 2px solid lightgreen">Tip</span>**?

## Related Issue(s)

It doesn't seem that there are any...


folke:
ty

PR #133, opened 2024-11-21
fix(gitbrowse): fix the regex in url_patterns

fang2hou:
## Description

The `remote` is used as a regex pattern, so the dot in the `"github.com"` means any character. This PR correct the regexes in default `url_patterns`.

https://github.com/folke/snacks.nvim/blob/aa38175c002555aded511b6787380ad887665680/lua/snacks/gitbrowse.lua#L40-L49

https://github.com/folke/snacks.nvim/blob/aa38175c002555aded511b6787380ad887665680/lua/snacks/gitbrowse.lua#L68-L72

Test code:

```lua
print(require("snacks").gitbrowse.get_url("https://githubacom.com/folke/snacks.nvim"))
```

## Screenshots

**Before**
image

**After**
image

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -38,11 +38,11 @@ local defaults = {
     { "%.git$"                            , "" },
   },
   url_patterns = {
-    ["github.com"] = {
+    ["github%.com"] = {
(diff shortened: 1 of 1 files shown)


folke:
ty!

PR #136, opened 2024-11-21
fix(git): recognize git worktrees as git repository

### LINKED ISSUE #115: bug: Snacks.git.get_root doesn't work for worktrees

- [X] I have updated the plugin to the latest version before submitting this issue
- [X] I have searched the existing issues of snacks.nvim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.10.1

### Operating system/version

CentOS8

### Describe the bug

https://github.com/folke/snacks.nvim/blob/198e4b95d58d0a2f027da61658221861a5b5d4f8/lua/snacks/git.lua#L22 checks if there exists a `.git` directory. Git worktrees have a `.git` file instead of a `.git` directory in the worktree directory and are not recognized as a git root directory by `Snacks.git`.

### Steps To Reproduce

1. git clone --bare -- <some_repo> <some_dir> && cd <some_dir>
2. git worktree add <main_branch> && cd <main_branch>
3. call `Snacks.git.get_root(<some_file_in_the_worktree>)`

### Expected Behavior

recognize a git worktree directory as a git repo. Replacing https://github.com/folke/snacks.nvim/blob/198e4b95d58d0a2f027da61658221861a5b5d4f8/lua/snacks/git.lua#L22 with:

`if vim.fn.isdirectory(dir .. "/.git") == 1 or vim.fn.filereadable(dir .. "/.git") == 1 then`

works, but not sure if this has any undesired side-effects.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

linusboehm:
## Description

`get_root` recognizes git worktrees as git repos

## Related Issue(s)

- Fixes #115

### DIFF
--- a/lua/snacks/git.lua
+++ b/lua/snacks/git.lua
@@ -20,7 +20,7 @@ function M.get_root(path)
   path = path == "" and (vim.uv or vim.loop).cwd() or path
   local git_root ---@type string?
   for dir in vim.fs.parents(path) do
-    if vim.uv.fs_stat(dir .. "/.git") ~= nil then
+    if vim.uv.fs_stat(dir .. "/.git") then
       git_root = dir
       break
     end
--- a/lua/snacks/git.lua
+++ b/lua/snacks/git.lua
@@ -20,7 +20,7 @@ function M.get_root(path)
   path = path == "" and (vim.uv or vim.loop).cwd() or path
   local git_root ---@type string?
   for dir in vim.fs.parents(path) do
-    if vim.fn.isdirectory(dir .. "/.git") == 1 or vim.fn.filereadable(dir .. "/.git") == 1 then
+    if vim.uv.fs_stat(dir .. "/.git") ~= nil then
       git_root = dir
       break
     end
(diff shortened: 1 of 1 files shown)


folke:
you could just do vim.fs_stat(...) instead of two checks.
reason I didnt' do this yet, is because vim.fn.isdirectory is about 10x faster and this matters for the recent projects on dashboards....

linusboehm:
I see. I wasn't aware of big perf difference. I updated it to use fs_stat.
If you'd rather stick with the more performant options, it would be great if it was possible to opt-in to have it work for worktrees as well (and let the user decide if they want to pay the perf cost).

Thanks for all the amazing plugins!!

lua/snacks/git.lua
```diff
@@ -20,7 +20,7 @@ function M.get_root(path)
   path = path == "" and (vim.uv or vim.loop).cwd() or path
   local git_root ---@type string?
   for dir in vim.fs.parents(path) do
-    if vim.fn.isdirectory(dir .. "/.git") == 1 then
+    if vim.uv.fs_stat(dir .. "/.git") ~= nil then
```
folke, line 23: you can remove the `~= nil` part here

linusboehm: done

PR #141, opened 2024-11-22
fix(dashboard): typo

ugur-a:


folke:
the docs are auto-generated, so you'll need to update the lua file instead

PR #151, opened 2024-11-23
feat(gitbrowse): add separate getremotes

dgmora:
## Description

Similarly to opening in github, I would like to be able to yank the current url. For that I thought that allowing to get the remotes would be useful. 

## Related Issue(s)

(not filed / tried) In the yank case at least for me, it'd be nicer to yank the file url without the line and in visual mode to yank the current line or range. I can yank the current file with this, but it's a bit inconvenient:

```lua
print(Snacks.gitbrowse._getremotes({ url_patterns = { ["github%.com"] = { file = "/blob/{branch}/{file}" } } })[1].url)
```

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -90,7 +90,13 @@ function M.open(opts)
 end
 
 ---@param opts? snacks.gitbrowse.Config
-function M._open(opts)
+function M.getremotes(opts)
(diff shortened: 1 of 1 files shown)


folke:
You can already do this:

```lua
Snacks.gitbrowse({ open = function(url) vim.fn.setreg("+", url) end })
```

PR #161, opened 2024-11-25
feat(gitbrowse): open commit when word is valid hash

### LINKED ISSUE #138: feature: GitBrowse opens commit hash
### Did you check the docs?

Hey there, love what you've done with this plugin. As a result, I've been thinking about all my git plugins and realized that I could replace my fugitive workflow with a combination of snacks.lazygit, snacks.gitbrowse, mini.git, and diff.nvim. But there's one feature that I've been missing whose implementation is pretty close to what snacks.gitbrowse already has implemented.

When browsing git logs using git blame line or diffview.nvim, I often want to view the commit on github or at least extract a link for sharing, usually of the form `https://github.com/{org}/{repo}/commit/{commit_hash}`.

One solution I've toyed with is when the cursor is on a line `commit {HASH}` at the top of some git log entry, `Snacks.gitbrowse.open()`, should open that commit. I don't know if there's a more elegant or general solution, and I'm certainly open to alternatives and better solutions :)

If this behavior seems pretty bespoke, some options API for adding user defined fields / what's and a way to provide such a field to `_open` would work for me. Beyond branch and file, I'd add `commit` and provide the commit hash via a function to parse the current line / context from my cursor or something similar.

### Additional context

_No response_

dpetka2001:
## Description
`gitbrowse` opens commit when word under cursor is valid commit hash. Please check the Lua pattern for the commit hash in the validation function I used as I'm not confident as to if it covers all the cases. I only tested with `diffview` file history and `gitsigns` blame buffer as I don't use other tools.

## Related Issue(s)
Closes #138.

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -19,7 +19,7 @@ local defaults = {
     end
     vim.ui.open(url)
   end,
-  ---@type "repo" | "branch" | "file"
+  ---@type "repo" | "branch" | "file" | "commit"
   what = "file", -- what to open. not all remotes support all types
   -- patterns to transform remotes to an actual URL
   -- stylua: ignore
@@ -41,10 +41,12 @@ local defaults = {
     ["github%.com"] = {
       branch = "/tree/{branch}",
       file = "/blob/{branch}/{file}#L{line}",
+      commit = "/commit/{commit}",
     },
     ["gitlab%.com"] = {
       branch = "/-/tree/{branch}",
       file = "/-/blob/{branch}/{file}#L{line}",
+      commit = "/-/commit/{commit}",
     },
   },
 }
@@ -84,6 +86,16 @@ local function system(cmd, err)
   return vim.split(vim.trim(proc), "\n")
 end
 
+---@param hash string
+---@return boolean
+local function is_valid_commit_hash(hash)
+  if not hash:match("^[a-fA-F0-9]+$") then
+    return false
+  end
+  system({ "git", "rev-parse", "--verify", hash }, "Invalid commit hash")
+  return true
+end
+
 ---@param opts? snacks.gitbrowse.Config
 function M.open(opts)
   pcall(M._open, opts) -- errors are handled with notifications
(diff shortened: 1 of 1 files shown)


dpetka2001:
I tried to change the logic of the initial commit in accordance to how it already is, so that we don't need the additional `if..else` check. Hopefully this is more preferable? Please do make any changes you deem necessary.

dpetka2001:
Another thing to mind is, maybe we should store the function call in a variable at the top like `local is_commit = is_valid_commit_hash(word)` and then use `is_commit` in the next steps, so that the function call doesn't happen in every step of `opts.what` evaluation?

krishna-bala:
One thing I noticed is that `git rev-parse --verify {hash}` is called from the current working directory, which is not necessarily the same as the "project root" identified by `:LazyRoot`. Would be nice if we could leverage `:LazyRoot`, but not sure how we can do that without making LazyVim a dependency of snacks.nvim.

krishna-bala:
Sorry I didn't explain the situation well. 

Let's say I open two buffers, `~/projectA/fileA.lua` and `~/projectB/fileB.lua`. 

I set my working directory to `~/projectA` (via `:chdir ~/projectA`). 

Scenario 1: 
When focused on the `~/projectA/fileA.lua` buffer, I call `:Gitsigns blame`. This opens a new `gitsigns-blame` window. I place my cursor on a commit hash in that `gitsigns-blame` window and call `:lua Snacks.gitbrowse()`. This opens the commit in my browser as expected.

Scenario 2: 
If I don't update the `cwd` and try the same workflow from the `~/projectB/fileB.lua` buffer (open a `gitsigns-blame`, place cursor on commit hash, call `:lua Snacks.gitbrowse()`), I get the "Invalid commit hash" error message.

---

My previous comment was more to say: it would be nice if `Snacks.gitbrowse()` was aware of the project context and didn't require the user to `:chdir ~/projectB` before calling `:lua Snacks.gitbrowse()` for scenario 2.

I tried making a change to the `system` call to use `require("lazyvim.util.root").get()` like so:

`system({ "git", "-C", require("lazyvim.util.root").get(), "rev-parse", "--verify", hash }, "Invalid commit hash")`

But that doesn't work because the `require("lazyvim.util.root").get()` is called from the `gitsigns-blame` window, which is a scratch window in the `cwd`.

dpetka2001:
I don't believe there's an easy way to do that. `gitsigns` and `diffview` are both not associated with any buffer. How can you determine the `git_root` directory without providing a file?

 I believe the most prudent thing would be to let the user change the `cwd` himself. You can easily create a mapping to `:cd expand("%:h")` for example.

dpetka2001:
@krishna-bala So, I managed to hack something to get around the `cwd` by using the alternate buffer, but I believe there are disadvantages as there might be quite a few edge cases. 

I managed to make it work with `gitsigns` and `diffview`, because both of them open in the correct repo of the tracked file when you invoke their corresponding commands. `neogit` also works if you invoke it with `Neogit cwd=%:p:h`, but I don't know how other tools would behave and I'm afraid there might be edge cases not currently taken into account.

In my personal opinion, I believe it would be better that the handling of the `cwd` would be left up to the user. 

@folke What do you think about that? Should I push my latest changes that take `cwd` of different repos into account or not? I don't expect an immediate answer. Whenever you have some free time. Just let me know if something like this should be handled internally by `gitbrowse` or left up to the user.

krishna-bala:
@dpetka2001 thanks for looking into this! I agree, seems reasonable to let the user handle changing `cwd` when necessary.

folke (changes requested):

lua/snacks/gitbrowse.lua
```diff
@@ -84,6 +86,16 @@ local function system(cmd, err)
   return vim.split(vim.trim(proc), "\n")
 end
 
+---@param hash string
+---@return boolean
+local function is_valid_commit_hash(hash)
+  if not hash:match("^[a-fA-F0-9]+$") then
+    return false
+  end
+  system({ "git", "rev-parse", "--verify", hash }, "Invalid commit hash")
```
folke, line 95: This should use the `-C cwd` param like the other checks

folke:
@dpetka2001 about that `cwd`, I wouldn't do anything special here.
Just use the cwd as detected for the other commands (which will result in the real cwd in most cases) and use that.

For sure not needed to handle opening git blame with commit hashes from another repo...

dpetka2001:
Done.

folke:
ty!

PR #172, opened 2024-11-29
feat(gitbrowse): allow custom branch

### LINKED ISSUE #170: feature: Override gitbrowse branch
### Did you check the docs?

When I want to copy or open the git url with gitbrowse, many times I actually want to copy the file from the `main`, not the current branch. I _think_ that's not possible with the current implementation

A way that it's easy to use gitbrowse with a specific branch.

I considered overwritting `open`, but my branches have `/`, so it's not very practical.

### Additional context

_No response_

dpetka2001:
## Description
Allow the user to be able to define custom branch with `Snacks.gitbrowse`. I also added `start_line` and `end_line` (default to `nil`), because from my testing when using a keymap such as the one in [my comment in the related issue](https://github.com/folke/snacks.nvim/issues/170#issuecomment-2508476110) or this one which I came up later while implementing this
```lua
vim.keymap.set({ "n", "x" }, "<leader>gY", function()
  local function system(cmd, err)
    local proc = vim.fn.system(cmd)
    if vim.v.shell_error ~= 0 then
      Snacks.notify.error({ err, proc }, { title = "Git Browse" })
      error(err)
    end
    return vim.split(vim.trim(proc), "\n")
  end
  local branches = {}
  local start_line, end_line
  if vim.fn.mode() == "v" or vim.fn.mode() == "V" then
    start_line = vim.fn.line("v")
    end_line = vim.fn.line(".")
    if start_line > end_line then
      start_line, end_line = end_line, start_line
    end
  end
  for _, branch in ipairs(system({ "git", "branch" }, "Failed to get git branches")) do
    branch = branch:match("%s*[%*]?%s*(%S+)")
    table.insert(branches, branch)
  end
  vim.ui.select(branches, {
    prompt = "Select branch",
  }, function(choice)
    if choice then
      Snacks.gitbrowse({ start_line = start_line or nil, end_line = end_line or nil, branch = choice })
    end
  end)
end, { desc = "Git Browse ()" })
```
Then the visual lines would not work. My guess is that in the first case the command line prompt and in the second case the `vim.ui.select` before calling `Snacks.gitbrowse` break the Visual mode state and any visual selection is not propagated to `Snacks.gitbrowse` to be properly evaluated by the internal code for visual lines selection, which works just fine when you just call directly `Snacks.gitbrowse()` via a function in a keymap (which I assume doesn't break Visual mode state). That's why I had to implement the visual selection again in the keymap as well and needed the extra parameters to pass the newly evaluated values.

Please do correct me if I'm somehow wrong in my logical deduction (which is based mostly on observation and guessing) and maybe `start_line` and `end_line` are not needed. Except if the user passes the option directly in the keymap, but then he would have to create multiple keymaps for different branches, which in my personal opinion defeats the purpose of this feature.

In both cases of the keymaps I mentioned the user can dynamically either type or choose the branch available to him. But maybe I'm missing something in both the keymaps implementations that I came up with.

## Related Issue(s)
Closes #170.


folke:
You need to exit visual mode first before you can get the selection marks.

See https://github.com/folke/snacks.nvim/blob/ddaa2aaba59bbd05c03992bfb295c98ccd3b3e50/lua/snacks/debug.lua#L53-L61 for example

folke:
I just fixed getting the visual selection, so your work-around should no longer be needed

folke:
merged. ty!

dpetka2001:
I'm not sure what you mean by my workaround should no longer be needed. But I still do have to calculate the visual selections in the keymap. If you try the keymap mentioned in my first comment, even with your recent changes it still behaves the same as I mentioned. So if you comment out the lines 
```lua
  local start_line, end_line
  if vim.fn.mode() == "v" or vim.fn.mode() == "V" then
    start_line = vim.fn.line("v")
    end_line = vim.fn.line(".")
    if start_line > end_line then
      start_line, end_line = end_line, start_line
    end
  end
```
which I suppose you mean is my workaround, then it opens the branch that you choose but with `vim.fn.line(".")`.
And I think it's due to using a `vim.ui.select` of my own that lists the current branches, that breaks the Visual mode state and then `Snacks.gitbrowse()` is called in the `function(choice)` of my own `vim.ui.select`. 

Now it still works fine, so I really don't have any problems with the implementation and since `opts.line_start` and `opts.line_end` did not get removed, so I can use them to pass my own values calculated in the keymap itself. I just thought I'd mention this, because I've not seen the implementation you redid before and am not familiar with it and honestly do not understand the logic of having to exit Visual mode to calculate the `>/<` marks (I've always used `v` and `.` because in `:h getpos()` it mentioned that it updated right away compared to `</>` in Visual mode). But from my testing it still behaves the same as the one that used to be if I comment out the Visual mode state recalculation in the keymap itself.

PR #186, opened 2024-12-01
feat(gitbrowse): add custom range formatter

skoch13:
## Description

This PR defines new option 

```lua
  ---@type table<string, fun(line_start: string, line_end: string): string?>
  range_formatters
```

that allows specifying a custom line range formatter for services that use a specific range format, e.g.

```lua
        range_formatters = {
          ["another%.git%.service"] = function(range_start, range_end)
            return range_end > range_start and range_start .. ("&lines-count=" .. range_end - range_start + 1) or ""
          end,
        },

```

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -6,6 +6,29 @@ local M = setmetatable({}, {
   end,
 })
 
+---@param line_start string
+---@param line_end string
+---@return string
+local function get_default_range_format(line_start, line_end)
+  return line_end > line_start and line_start .. "-L" .. line_end or line_start
+end
+
+---@param remote string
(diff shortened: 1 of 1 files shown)


folke:
I changed the code to expose `line_start` and `line_end` directly, which should cover your use-case.

skoch13:
@folke thank you,
I cannot figure out how interpolation would work here:

I'm trying to specify it as follows 

```lua
file = "/files/{branch}/{file}?&line={line_start}"
  .. "{line_end} > {line_start} and ('&lines-count=' .. {line_end} - {line_start} + 1) or ''}",
```

but it doesn't work

folke:
Just added `line_count`, so you can use that directly

PR #196, opened 2024-12-03
feat(snacks): zen mode

folke:
## Description

A bunch of new plugins:
- animate: fast & efficient animations with 45 included easing functions (library)
- scope: scope detection based on treesitter or indents (library)
- indent: fast indent guides and scope highlighting
  - uses window-local extmarks (= decoration provider + ephemeral extmarks)
- dim: better twilight.nvim (uses scope)
  - window-local extmarks  
- zen: better zen mode and integrates with toggle
  - zoom mode expands the current buffer like there would be no splits
  - exiting the zen/zoom window closes the zen/zoom window
  - when the zen/zoom window buffer changes, so does the buffer in the original window
  - Zoom with `lua Snacks.zen.zoom()`

## Testing

```lua
return {
	{ "indent-blankline.nvim", enabled = false },
	{
		"snacks.nvim",
		opts = {
			indent = {},
		},
    -- stylua: ignore
    keys = {
      { "<leader>z", function() Snacks.zen() end, desc = "Toggle Zen Mode" },
      { "<leader>Z", function() Snacks.zen.zoom() end, desc = "Toggle Zoom" },
    },
	},
	{
		"snacks.nvim",
		opts = function()
			LazyVim.on_load("which-key.nvim", function()
				Snacks.toggle.indent():map("<leader>ug")
				Snacks.toggle.dim():map("<leader>uD")
			end)
		end,
	},
}
```

## `lua Snacks.zen.zen()`

image

## `lua Snacks.zen.zen({backdrop = {transparent = false, blend = 70}})`

image


vdegenne:
Not sure if that's me having a very specific requirement for dimming but the defaults (using treesitter) look incorrect, I know it's not a bug, more like the treesitter for js/ts losing itself in complex multi-language files.   

In this video I show how the defaults look like and then I disable the treesitter to demonstrate that the dimming look more like what one would expect:


https://github.com/user-attachments/assets/b11e814f-fae3-4db8-808d-4c8e3bdc5920

But it's still not perfect (apart from the fact that the treesitter is not disabled...)

I still think you should give an option for using to choose the dimming strategy:

- treesitter (actually the default)
- indentscope (how it behaves when the treesitter is disabled, dim only the code between the vertical bright indent line)
- indentscope-cursor (same as indentscope but relative to cursor position)

In the following video, I use "indentscope-cursor" strategy (using `mini-indentscope` + `twilight`):


https://github.com/user-attachments/assets/976e5688-6848-4ac3-8a69-cd0c99ee976a


imho, *at least* `treesitter` and `indentscope` should be two options to choose from, to be able to dim the current indentation while retaining treesitter capabilities.

folke:
The whole dimming is highly configurable. There is no **correct** way of dimming.
Wait till I finished the docs.

folke:
btw, I'm reworking the whole scroll plugin, so better not to use it for now.

vdegenne:
> btw, I'm reworking the whole scroll plugin, so better not to use it for now.

When I tried to use it it wasn't working at all :smile:

folke:
The scroll plugin should work as it should now :)
It also properly deals with mousescroll. That's something I had to hack in mini.animate before in LAzyVim.

If you want to test the scroll plugin, make sure to disable mini.animate, or at least its scrolling module.

Would love to hear some feedback on this plugin, since it's pretty tricky to get it work under all circumstances.

It should pretty much work exactly the same now as if it would not be enabled. Things like scrolloff, mousescroll etc should all work correctly.

dpetka2001:
I've been using it since yesterday with default settings and the only problem I had found was the mouse scrolling. After latest pull this also works as expected. I just increased `step` to 20 for smoother transition for my own preferences. If I find anything weird I will report back.

vdegenne:
Really like the scroll feature, and the default speed is good. I just had this problem, If I have 

```lua
opts = {
  scroll = { enabled = false }
},
init = function()
  vim.api.nvim_create_autocmd("User", {
    pattern = "VeryLazy",
    callback = function()
      Snacks.toggle.scroll():map("<leader>uS")
    end,
  })
end
```

And restart nvim, the scroll feature is off as expected but but pressing `<leader>u` and I get this odd state:

image

iow, snack thinks the feature is on.. pressing `S` will disable it even though it's already disabled.

dpetka2001:
That's just disabling the plugin. The toggle for state checks the variable `vim.g.snacks_scroll` for the state. So, set that to `false` if you want it disabled by default and then it will get the correct state.

vdegenne:
@dpetka2001 What I mean is that, isn't Snacks supposed to do this checking on startup and set the toggle accordingly whether the feature is on or off?

vdegenne:
I understand yes but in the example I provided Snacks "thinks" it's activated. If the module is not loaded can it just set it to false by default?

dpetka2001:
You should just guard your mapping with
```lua
if LazyVim.opts("snacks.nvim").scroll ~= nil and LazyVim.opts("snacks.nvim").scroll.enabled ~= false then
  Snacks.toggle.scroll():map("<leader>uS")
end
```
That way your mapping will not be available when the `scroll` snack is not enabled. It can't read the state if the snack is not loaded.

vdegenne:
@folke When using 
```lua
terminal = {
  win = {
    position = "top" -- or any other value except "float"
```

lazygit will open in specified position:

image

LazyGit should always open afloat I think

folke:
Use this instead: `styles = { terminal = { position = "top" } }`.
Otherwise you are indeed overriding the options for all terminals, including lazygit.
lazygit has its own style you can change as well.

vdegenne:
Last suggestion of the day:
when renaming a file, can the cursor be before the extension at start?

image

It's more rare to change extension of a file, and using arrow keys to go left when renaming (because there is no normal mode available in the input) does not feel right in vim.

folke:
Nope, `vim.ui.input` does not have a way to specify the position.

vdegenne:
Too bad. That's a recent update? I haven't noticed when it changed but when I was using `<leader>gc` to rename objects in my code, the input had normal mode capability, could use `h` and `l` to travel inside the input, could even use `gu` or `gU` a selection. It was superior imo.

folke:
You'll be able to do that soon. Finishing my `snacks.input` plugin. Forgot I had most of that ready :)

vdegenne:
I don't know if that's practical but I share this idea,

```lua
Snacks.toggle.dim():map('<leader>uD'):persist()
```

So the toggle will persist upon restarts.

PR #201, opened 2024-12-04
docs(lazygit): comment how to disable nerd icons

### LINKED ISSUE #200: feature(lazygit): disable nerd icons
### Did you check the docs?

Currently the lazygit module has default `nerdFontsVersion = "3"`, setting this to other number causes lazygit to exit with non zero exit code as it is a non supported font version.

It would be great to have an option to disable icons in the lazygit window. Perhaps a separate option to ignore nerd icon fonts.

I have tried the following
* setting `nerdFontsVersion` to other string values which cause lazygit to error out
* setting `nerdFontsVersion` to `false`, causes snacks to fail with an error `attempting to concatenate a non string value
* setting `nerdFontsVersion` to `nil` is ignored as the table ignores this value and the `snacks.lazygit` default value `"3"` is used

### Additional context

I was able to remove nerd icons by commenting out `nerdFontsVersion = "3"` in the snacks.lazygit source file

PS: Thank you for your work!

antonk52:
## Description

Show explicitly how to disable nerd icons in the default versions of config and documentation

## Related Issue(s)

- Fixes #200

### DIFF
--- a/docs/lazygit.md
+++ b/docs/lazygit.md
@@ -38,7 +38,10 @@ and integrate edit with the current neovim instance.
   -- you need to double quote it: `"\"test\""`
   config = {
     os = { editPreset = "nvim-remote" },
-    gui = { nerdFontsVersion = "3" },
+    gui = {
+      -- set to an empty string "" to disable icons
+      nerdFontsVersion = "3",
+    },
   },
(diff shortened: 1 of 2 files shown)


folke:
ty

PR #208, opened 2024-12-07
feat(lazygit): respect existing LG_CONFIG_FILE when setting config paths

sswensen:
## Description

This change preserves any existing Lazygit user config while still adding the theme configuration.

New Behavior:
1. Check if `LG_CONFIG_FILE` already exists in the environment (`vim.env.LG_CONFIG_FILE`)
2. If it exists, use that value; if not, fall back to the default config path (`vim.fs.normalize(config_dir .. "/config.yml")`)
3. Append the normalized theme configuration path (`opts.theme_path`)


lua/snacks/lazygit.lua
```diff
@@ -77,7 +77,7 @@ local function env(opts)
 
     if vim.v.shell_error == 0 and #lines > 1 then
       config_dir = vim.split(lines[1], "\n", { plain = true })[1]
-      vim.env.LG_CONFIG_FILE = vim.fs.normalize(config_dir .. "/config.yml" .. "," .. opts.theme_path)
+      vim.env.LG_CONFIG_FILE = (vim.env.LG_CONFIG_FILE or vim.fs.normalize(config_dir .. "/config.yml")) .. "," .. opts.theme_path
```
folke, line 80: This will keep extending that env var, so won't work as expected.
Add a module level `LG_CONFIG_FILE = vim.env.LG_CONFIG_FILE` once and use that instead so we don't keep extending it with the theme path every time lazygit is laucnhed

folke (changes requested):

lua/snacks/lazygit.lua
sswensen: Moved env lookup to outside of `env` function. Resolves repeated appending of theme config

sswensen:
@folke Any other updates needed to this?
FWIW I was unable to force the appending of the theme config to `vim.env.LG_CONFIG_FILE` more than once, likely due to the `config_dir` var preventing the env code from running on subsequent Lazygit opens

folke:
Still need to check. This PR actually probably breaks my own setup, where I use LG_CONFIG_FILE to point to the snacks generated config files. This way I can use the snacks config outside of Neovim.
With this PR, in its current state, it could give issues.

Will check tomorrow and probably make some adjustments.

PR #215, opened 2024-12-07
fix(dashboard): add filetype to terminal sections

pynappo:
The intention is to make it easier to exclude the dashboard terminal from some plugins. For example, heirline.nvim sets winbar per-window to allow for winbar to be optionally disabled in non-floating windows, so this PR lets me disable heirline for snacks' dashboard terminals over other floating terminals.


folke:
ty!

PR #292, opened 2024-12-14
fix(input): missing padding if neither title nor icon positioned left

chrisgrieser:
## Description
The new`icon_pos` /`title_pos` options resulted in a situation where there is missing padding when neither is positioned left. This PR fixes this by falling back to `" "` as statuscolumn.

## Screenshots
before
before

after
after


folke:
Right, didn't think about that. ty!

PR #295, opened 2024-12-14
feat(notifier): `Telescope notifier`

kawre:
## Description

This is a port of https://github.com/rcarriga/nvim-notify?tab=readme-ov-file#viewing-history for the snacks notifier.
It simplifies navigating through notifications, as opposed to having them dumped into a single large file.

## Screenshots

https://github.com/user-attachments/assets/4969342c-c933-4295-b8ac-9e32e8bc93e8


justgook:
btw, how do you done that cmdline

folke:
`:lua Snacks.picker.notifications()`

PR #297, opened 2024-12-14
feat(notifier): `hide_from_history` option

kawre:
## Description

Option to exclude some notifications from notification history,
for example, when dealing with spinners.

### DIFF
--- a/lua/snacks/notifier.lua
+++ b/lua/snacks/notifier.lua
@@ -34,7 +34,7 @@ local uv = vim.uv or vim.loop
 ---@field style? snacks.notifier.style
 ---@field opts? fun(notif: snacks.notifier.Notif) -- dynamic opts
 ---@field hl? snacks.notifier.hl -- highlight overrides
----@field hide_from_history? boolean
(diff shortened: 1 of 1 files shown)


lua/snacks/notifier.lua
```diff
@@ -34,6 +34,7 @@ local uv = vim.uv or vim.loop
 ---@field style? snacks.notifier.style
 ---@field opts? fun(notif: snacks.notifier.Notif) -- dynamic opts
 ---@field hl? snacks.notifier.hl -- highlight overrides
+---@field hide_from_history? boolean
```
folke, line 37: Can yo change it to `history`? And then check for `opts.history ~= false`?

kawre: nvim-notify already uses `hide_from_history`, should i check for both `history` and `hide_from_history`?

folke: nope, not needed to be compatible with `nvim-notify`

folke:
ty!

PR #306, opened 2024-12-15
feat(styles): exportable styles

vdegenne:
To allow syntax like:

```lua
local mystyle = Snacks.config.style('...')

-- somewhere else
Snacks.win({
  style = mystyle
})
```

and improve code navigation (e.g. jumping using `gd`, as opposed to searching string, or not being sure of the exact name, ...)


folke:
Makes sense. ty!

PR #307, opened 2024-12-15
fix(notifier): open history window with correct style

dpetka2001:
## Description
This [commit](https://github.com/folke/snacks.nvim/commit/fd9ef30206185e3dd4d3294c74e2fd0dee9722d1) renamed the style for `history_notification`, but opening the window still used the old name.


folke:
Ty

PR #311, opened 2024-12-16
chore(docs): updated gitbrowse keymap to support visual mode

lewisdenny:
## Description

Since https://github.com/folke/snacks.nvim/pull/89 was merged gitbrowse supports opening multiline URLs. 
This patch updates the keymap example docs to include visual mode so the feature works as expected.


folke:
That file is autogenerated. Please update docs/examples/init.lua instead

lewisdenny:
Done, thanks mate.

folke:
ty!

PR #328, opened 2024-12-17
fix(input): use correct highlight group for input prompt

willnorris:
I'm suspect this was just a typo... using the border highlight group rather than title.


willnorris:
or maybe it should only use the title highlight group when `prompt_pos == "title"` ?  I'm not sure what other values prompt_pos can have and whether using the border group makes sense in other instances.

folke:
Should actually be SnacksInputPrompt. For sure a typo

folke:
ty

PR #377, opened 2024-12-23
fix(lazygit): enable boolean values in config

NevLext:
## Description

Using boolean value inside `config` resulted in error `attempt to concatenate a boolean value` inside `to_yaml` function.
This PR fixes it by converting boolean values to string

Example config:
```
lazygit = {
    config = {
        gui = {
            showFileTree = false,
        },
    },
},
```
## Screenshots
image


folke:
ty!

PR #378, opened 2024-12-23
fix(toggle): only execute `self:_wk` when mapping exists

dpetka2001:
## Description
Currently when a user disables a key with `vim.keymap.del` in his personal configuration, the `which-key` menu still shows the mapping. This ensures that `self:_wk` function won't execute if the mapping doesn't exist.

If you're satisfied with this solution, please also remove the comment I included.

## Related Issue(s)
No issues, rather a [LazyVim discussion](https://github.com/LazyVim/LazyVim/discussions/5240)

## Screenshots

**Before**

**After**

### DIFF
--- a/lua/snacks/toggle.lua
+++ b/lua/snacks/toggle.lua
@@ -116,8 +116,19 @@ function Toggle:map(keys, opts)
     self:toggle()
   end, opts)
   if self.opts.which_key then
-    Snacks.util.on_module("which-key", function()
-      self:_wk(keys, mode)
+    -- Needed so that it executes after `keymaps.lua` loads, otherwise it
+    -- wouldn't work as `Snacks.toggle.animate():map("<leader>ua") will be
+    -- evaluated during the parsing phase. Maybe there's a better way??
+    vim.schedule(function()
+      Snacks.util.on_module("which-key", function()
+        local modes = type(mode) == "string" and { mode } or mode
+        for _, m in ipairs(modes) do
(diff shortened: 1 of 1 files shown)


folke:
While this works, this is really a work-around for something specific to LazyVim, so it doesn't really make sense to add this in snacks.

dpetka2001:
This is a workaround for any distro that might use `snacks.toggle:map` to define toggleable keys except for people who do their own custom config and have full control over it. The only specific LazyVim workaround is the `vim.schedule` which is needed because `keymaps` might or might not get lazy-loaded in LazyVim depending on how you invoke Neovim.

However, this is entirely your decision and doesn't affect me to be honest, so I really don't mind. I only made this for a past discussion that was present in LazyVim and only updated it today because of the current LazyVim issue.

folke:
I fixed it in which-key. Added a new `mapping.real` property that when set and the real keymap does not exists, then it gets hidden.

dpetka2001:
Cool, I will reference your comment in today's LazyVim issue, so that OP is aware of the new possibility.

dpetka2001:
And you also integrated it into `Toggle._wk()` as well. Really nice. At least it saves users trouble having to remap their mappings in which-key `spec` with `real = true`. 

Again, this doesn't affect me regardless of what you would decide. I'm too used to LazyVim's mappings as they currently are after having used it for almost 2 years.

folke:
should be all automatic now. so all good :)

Issue #379, opened 2024-12-23
bug: snacks dashboard terminal section caching throws error when handling long commands


Jay-Madden:
### Did you check docs and existing issues?

- [X] I have read all the snacks.nvim docs
- [X] I have updated the plugin to the latest version before submitting this issue
- [X] I have searched the existing issues of snacks.nvim
- [X] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.0

### Operating system/version

macOS Sonoma Version 14.5 (23F79)

### Describe the bug

When attempting to pass a long command or script to a `terminal` section. The automatic caching code will throw an error if the script passed in is too long. 

Error message: 
```
[...]

PR #381, opened 2024-12-23
fix(dashboard): hash dashboard terminal section caching key to support long commands

Jay-Madden:
## Description

This change hashes the caching key for snacks dashboards terminal sections instead of building the key from the command itself. This allows for arbitrary length scripts to be given as a terminal section command instead of relying on existing commands in PATH. 

## Related Issue(s)

#379 

## Screenshots

image


folke:
That's awesome, thanks!
TIL about `vim.fn.sha256`

PR #400, opened 2024-12-27
fix(win): backdrop having bright cell at top right

chrisgrieser:
## The bug
`backdrop` displays a bright cell at the top left. upon investigation, I figured out that it's created by the user's `colorcolumn` setting also being applied to the backdrop.

## The fix
Disable `colorcolumn` for the backdrop window.

## Screenshots

before
Pasted image 2024-12-27 at 14 16 40@2x

after
Pasted image 2024-12-27 at 14 17 14@2x


folke:
ty!

PR #413, opened 2025-01-01
fix(win): window position with border_offset logic

hareki:
## Description

The window position is not correctly centered when position is set to `float` and border isn't `none`. I have fixed this issue once [in the past](https://github.com/folke/snacks.nvim/pull/64) but it seems to have returned after the [dimension calculation rework](https://github.com/folke/snacks.nvim/commit/cc0b52872b99e3af7d80536e8a9cbc28d47e7f19). So I have applied the same logic to the new structure.

Thank you for your work and happy new year!

### Before
before

### After
after


folke:
Had to do this a little bit different to make it work for all cases.
Should be fixed now. ty!

PR #425, opened 2025-01-06
fix(debug.run): schedule only nvim_buf_set_extmark in on_print

stefanboca:
## Description

Previously, `get_line` was called from a scheduled function, and would always return nil, placing the extmark for print output on the wrong line.

## Screenshots

Before:
image

After:
image


folke:
ty

PR #427, opened 2025-01-06
fix(win): incorrect negative position calculation when relative to cursor

amedoeyes:
When making a window relative to the cursor and setting either `row` or `col` to a negative integer it moves relative to the bottom right of its parent instead of moving relative to the cursor.

Example with input:
```lua
styles = {
  input = {
    relative = "cursor",
    width = 40,
    row = -3,
    col = 0,
    border = "single",
    title_pos = "left",
    },
},
```
Window opens like this:
image
When it should be like this:
image

### DIFF
--- a/lua/snacks/win.lua
+++ b/lua/snacks/win.lua
@@ -876,15 +876,18 @@ function M:dim(parent)
   ---@param s number size
   ---@param ps number parent size
   local function pos(p, s, ps, border_from, border_to)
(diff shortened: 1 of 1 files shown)


folke:
Thank you for the PR. Changed to properly fix it in all conditions.

folke:
Yes, you're right. It should only do the fn part. Just fixed that.

PR #445, opened 2025-01-10
feat(snacks): added `snacks.picker`

folke:
## Description

More info coming tomorrow.

In short:
- very fast. pretty much realtime filtering/sorting in huge repos (like 1.7 million files)
- extensible
- easy to customize the layout (and lots of presets) with `snacks.layout`
- simple to create custom pickers
- `vim.ui.select`
- lots of builtin pickers
- uses treesitter highlighting wherever it makes sense
- fast lua fuzzy matcher which supports the [fzf syntax](https://junegunn.github.io/fzf/search-syntax/) and additionally supports field filters, like `file:lua$ 'function`

There's no snacks picker command, just use lua.

```lua
-- all pickers
Snacks.picker()

-- run files picker
Snacks.picker.files(opts)
Snacks.picker.pick("files", opts)
Snacks.picker.pick({source = "files", ...})
```

## Todo
- [x] issue with preview loc not always correct when scrolling fast in list (probably due to `snacks.scroll`)
- [x] `grep` (`live_grep`) is sometimes too fast in large repos and can impact ui rendering. Not very noticeable, but something I want to look at.
- [x] docs
- [x] treesitter highlights are broken. Messed something up somewhere


folke:
@mathjiajia yeah, noticed as well. The docs are still a work in progress. Will get to it now! :)

milanglacier:
This looks awesome. 

Just be curious, why the picker is not published as a standalone plugin? Seems like the feature set and LoC of the picker are more like a "dinner" rather than a "snack".

folke:
Why?

PR #471, opened 2025-01-15
fix(picker): select_and_prev should use list_up instead of list_down

linnovs:
## Description

Fix the picker select_and_prev functionality.


folke:
ty!

PR #500, opened 2025-01-15
feat(picker): add git file diff previewer

### LINKED ISSUE #478: feature: Git status previewer similar to Telescope
### Did you check the docs?

Git status preview uses the default git pager. Outside of nvim Im' using `diff-so-fancy`, which cannot be scrolled up/down through Snacks picker.

I'd prefer a git status previewer similar or same to the Telescope ones.
Telescope creates a diff for a file and shows that as a buffer preview with a filetype `diff`.

Example:
Telescope:
screenshot_2025_01_15_10_17_45

Snacks (with diff-so-fancy):
screenshot_2025_01_15_10_18_14

Creating my own picker, but I wouldn't go down that route if not necessary.

### Additional context

_No response_

kristijanhusak:
## Description

Add a `git_file_diff` previewer to preview git diff/status in a same way as Telescope.
Test it like this:
```lua
Snacks.picker.git_status({
  preview = 'git_file_diff'
})
```

## Related Issue(s)

- Fixes #478

## Screenshots

Before: 
screenshot_2025_01_15_17_16_49

After:

screenshot_2025_01_15_17_16_26

### DIFF
--- a/lua/snacks/picker/preview.lua
+++ b/lua/snacks/picker/preview.lua
@@ -205,6 +205,44 @@ function M.git_diff(ctx)
 end
 
 ---@param ctx snacks.picker.preview.ctx
+function M.git_file_diff(ctx)
+  ctx.preview:reset()
+  local output = {}
+  local killed = false
+  local jid = vim.fn.jobstart({ "git", "--no-pager", "diff", "HEAD", "--", ctx.item.file }, {
+    on_stdout = function(_, data)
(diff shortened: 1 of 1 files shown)


kristijanhusak:
We should probably disable line numbers by default. Please let me know how to do that correctly. I'm assuming it should be not hardcoded.

folke:
Changed it to make it configurable to use git native or neovim preview.
Also re-used the `cmd` preview for this.
Default is now Neovim diff.

PR #517, opened 2025-01-16
feat(picker): support something like telescopes __locations_input option for file pickers #496

### LINKED ISSUE #496: feature(picker): support something like telescopes __locations_input option for file pickers
### Did you check the docs?

Hey there and thanks for this great collection of plugin.
I was overjoyed today when I saw that it now includes a picker and tried replacing telescope with it.
It works great for me right now except one very small-ish nitpick: Telescope added an option for file pickers to ignore/forward `:line:column` information at the end of filenames (see https://github.com/nvim-telescope/telescope.nvim/pull/2791).

I oftentimes have the usecase that i copy a filename e.g. from a ci run including this info and than pasting it into my file picker.
Currently I have to manually remove the info and the file will than be opened not at the correct line.

Something similar to how telescope does it, ignore the line/column information for fetching but forward it when opening a file to open it in the proper location.

I guess... stay with telescope? :D

### Additional context

_No response_

gcollura:
## Description

Pass file line and column from input to file match. Highlight that line as well.

I'm not sure if this is the best approach, but it works just fine in the most common use cases.

## Related Issue(s)

- Fixes #496

## Screenshots

image

### DIFF
--- a/lua/snacks/picker/core/filter.lua
+++ b/lua/snacks/picker/core/filter.lua
@@ -1,6 +1,7 @@
 ---@class snacks.picker.Filter
 ---@field pattern string Pattern used to filter items by the matcher
 ---@field search string Initial search string used by finders
+---@field loc? {[1]: number, [2]: number} Target location for the cursor
 ---@field buf? number
 ---@field file? string
 ---@field cwd string
--- a/lua/snacks/picker/core/input.lua
+++ b/lua/snacks/picker/core/input.lua
@@ -1,3 +1,4 @@
+local util = require("snacks.picker.util")
 ---@class snacks.picker.input
 ---@field win snacks.win
 ---@field totals string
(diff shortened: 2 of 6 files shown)


folke:
I did it a little different. See commit.

The part before the `:` should either contain a `.`, `/` or `\`.
This is to prevent issues with field searches.

When a file position is detected it is converted to a `file:filename$` search in the matcher.
positions then get automatically added (if the item has no position)

gcollura:
> I did it a little different. See commit.
> 
> 
> 
> The part before the `:` should either contain a `.`, `/` or `\`.
> 
> This is to prevent issues with field searches.
> 
> 
> 
> When a file position is detected it is converted to a `file:filename$` search in the matcher.
> 
> positions then get automatically added (if the item has no position)

Beautiful, thank you! I'll try it out later today

PR #519, opened 2025-01-16
feat(picker.git_diff): pick diff hunks

antonk52:
Hi and thank you for the useful plugin!

## Description

I have been exploring the new picker api and got this git diff picker working which I think can be a nice addition to the existing set of picker. The key difference from the existing git_status picker is that it separates each hunk into its own item similar to `git add --patch` and when you select an item it takes you to the first hunk's line instead of top of the buffer. Before I put more effort into it I would like to see if this is a welcome change first.

This implementation can be further extended by allowing to support additional arguments such as `--staged`, `--ignore-all-space` or both.

A few points on current implementations. I could not get the per line parsing to work using the `require("snacks.picker.source.proc").proc(...)` and fell back to manually parsing the output after the command finished executing. There are likely many ways to improve it. I have been using it with telescope for about a year and so far it worked great. I think more people can benefit from this picker.

## Related Issue(s)

none

## Screenshots

Demo

https://github.com/user-attachments/assets/6cdfa48e-bc29-4194-8430-092fbc5f3dcd

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -19,6 +19,14 @@ local M = {}
 ---@field smartcase? boolean use smartcase (defaults to true)
 ---@field ignorecase? boolean use ignorecase (defaults to true)
 
+--- This is only used when using `opts.preview = "preview"`.
+--- It's a previewer that shows a preview based on the item data.
+---@class snacks.picker.Item.preview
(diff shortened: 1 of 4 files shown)


folke:
Looks nice!
I'll fix the impl though, since you're creating a picker in the finder.

folke:
Refactored and merged. Thank you!

PR #546, opened 2025-01-16
Input: put cursor right after the default prompt

iserpent:
## Description

When the input opens, the cursor is placed two columns to the right after the default prompt, but should be one.
The cause is the `vim.api.nvim_win_set_cursor` takes (row, col) with the base of (1, 0) */I hate this/*

## Related Issue(s)

No related issue created.

## Screenshots

`:lua Snacks.input({ default = "aaa" }, function () end )`
gives the input as "aaa |" (where pipe shows the cursor position)


iserpent:
Will reopen with the commit message without "Input: ..." :)

PR #560, opened 2025-01-16
fix(picker): remove filename from lsp picker text

kristijanhusak:
## Description

Moving lsp pickers to qflist adds an unnecessary absolute path to the entries.
There's more details on discussion https://github.com/folke/snacks.nvim/discussions/536

## Related Issue(s)

No issue, but discussion #536


folke:
This is not correct.
An item's `text` property is what's being used for filtering, so by removing it you could no longer filter on the filename.

folke:
Yep, but need to think about it

PR #561, opened 2025-01-16
fix(bigfile): check existence of NoMatchParen before executing

ng-tony:
## Description

In my config I use the following flag to disable the default `MatchParen` plugin.

```
vim.g.loaded_matchparen = 1
```

This causes an error when `bigfile` tries to call the `NoMatchParen` command since it does not exist.
This PR fixes that by checking for the existence of the command before executing it.

### DIFF
--- a/lua/snacks/bigfile.lua
+++ b/lua/snacks/bigfile.lua
@@ -15,7 +15,9 @@ local defaults = {
   -- Enable or disable features when big file detected
   ---@param ctx {buf: number, ft:string}
   setup = function(ctx)
-    vim.cmd([[NoMatchParen]])
+    if vim.fn.exists("NoMatchParen") then
+      vim.cmd([[NoMatchParen]])
+    end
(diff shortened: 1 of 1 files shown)


lua/snacks/bigfile.lua
```diff
@@ -15,7 +15,9 @@ local defaults = {
   -- Enable or disable features when big file detected
   ---@param ctx {buf: number, ft:string}
   setup = function(ctx)
-    vim.cmd([[NoMatchParen]])
+    if vim.fn.exists("NoMatchParen") then
```
folke, line 18: exists probably returns 1 or 0? So this condition will always be true?

ng-tony: Yea you're right. Although in testing I ended up with vim.fn.exists(":NoMatchParen") evaluating to `2`

folke:
ty

PR #570, opened 2025-01-17
feat(statuscolumn): add filter for buffers to disable statuscolumn

emxxjnm:
## Description

 filter for buffers to disable statuscolumn

### DIFF
--- a/lua/snacks/statuscolumn.lua
+++ b/lua/snacks/statuscolumn.lua
@@ -26,6 +26,7 @@ local LINE_NR = "%=%{%(&number || &relativenumber) && v:virtnum == 0 ? ("
 ---@field left snacks.statuscolumn.Components
 ---@field right snacks.statuscolumn.Components
 ---@field enabled? boolean
+---@field filter? fun(buf:number):boolean
 local defaults = {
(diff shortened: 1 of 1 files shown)


folke:
What's the use-case here?
The statuscolumn will already be disabled in `neo-tree`

folke:
And the correct way to disable the statuscolumn is by setting `signcolumn = "no"` on the window.

emxxjnm:
As shown in the screenshot below, there is an extra blank placeholder before all lines
image

Set statuscolumn="" to remove blank placeholders, as shown below
image

emxxjnm:
> What's the use-case here? The statuscolumn will already be disabled in `neo-tree`

@folke Thanks for such a great plugin!

As shown in the screenshot below, there is an extra blank placeholder before all lines
image

Set statuscolumn="" to remove blank placeholders, as shown below
image

dpetka2001:
statuscolumn is indeed enabled in neo-tree and signcolumn is set to `no` already. Setting `statuscolumn = ""` does indeed remove that extra space. It can also be achieved in a `FileType` autocmd though if someone wants to remove it.

folke:
Just fixed that extra space

PR #574, opened 2025-01-17
fix(statuscolumn): use extmarks cursorline hl

minusfive:
## Description

Statuscolumn snack swallows the `cursorline_hl_group` on extmarks, causing defined `cursorline` highlights to break.

## Related Issue(s)

TBD (searching...)

## Screenshots

With appropriate GitSigns highlights defined on theme, e.g.:

```lua
-- plugin spec
{
  "lewis6991/gitsigns.nvim",
  opts = {
    culhl = true,
  },
},

--- Catppuccin
{
  GitSignsAddCul = { bg = colors.mantle, fg = colors.green },
  GitSignsChangeCul = { bg = colors.mantle, fg = colors.yellow },
  GitSignsDeleteCul = { bg = colors.mantle, fg = colors.red },
}
```

### Before
Screenshot 2025-01-17 at 01 34 56

### After
Screenshot 2025-01-17 at 01 35 06

### DIFF
--- a/lua/snacks/statuscolumn.lua
+++ b/lua/snacks/statuscolumn.lua
@@ -44,7 +44,7 @@ local config = Snacks.config.get("statuscolumn", defaults)
 
 ---@private
 ---@alias snacks.statuscolumn.Sign.type "mark"|"sign"|"fold"|"git"
----@alias snacks.statuscolumn.Sign {name:string, text:string, texthl:string, priority:number, type:snacks.statuscolumn.Sign.type}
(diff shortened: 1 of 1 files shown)


folke:
i dont think this is correct. Won't work if both a texthl and culhl are defined.
It may work in the gitsigns case, since they probably set the fg for the culhl to the same value.
The correct way would be to apply both hl groups, but I don't think that's possible with the `statuscolumn` syntax.

minusfive:
> i dont think this is correct. Won't work if both a texthl and culhl are defined.
> 
> It may work in the gitsigns case, since they probably set the fg for the culhl to the same value.
> 
> The correct way would be to apply both hl groups, but I don't think that's possible with the `statuscolumn` syntax.

🤔 Will keep poking at it later.

Issue #372, opened 2024-12-22
feature: configure bigfile based on max column length or path


cngu:
### Did you check the docs?

- [X] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

Consider minified JS/CSS files that consist of a few lines, but each line is thousands of columns long, amounting to 100KB.  Loading these minified files (with LSP, highlighting, and more enabled) take much longer than a regular 100 KB file split into a hundred lines.  Therefore enabling bigfile based solely on file size is insufficient - it needs to consider line length.

### Describe the solution you'd like

Enable the bigfile filetype if the max column length exceeds a configurable number.  It may be slow to find and count the max line length, in which case we could also allow configuring the maximum number of lines to search from the top.  Note that we can't always assume the first line will be long (e.g. it could be a short license comment with the long minified lines further down).

### Describe alternatives you've considered

Another option is to enable this based on a match on the absolute file path.  It'd be up to the user to find a common pattern match for their minified files (e.g. everything under a "dist/" directory, or file names that match "*.min.js").

disrupted:
I am still interested in this

folke:
bigfile now checks average line length (default 1000).

PR #576, opened 2025-01-17
feat(bigfile): take into account lines length

chrisands:
Added check for long lines
#372

### DIFF
--- a/doc/snacks-bigfile.txt
+++ b/doc/snacks-bigfile.txt
@@ -44,7 +44,8 @@ mini.animate <https://github.com/echasnovski/mini.animate> (if used)
     {
       notify = true, -- show notification when big file detected
       size = 1.5 * 1024 * 1024, -- 1.5MB
-      columns = 500,
+      line_length = 500,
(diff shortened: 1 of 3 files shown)


folke:
Sorry, but this is a big no for performance.
I'm willing to accept a PR that checks the length of the first buffer line `vim.api.nvim_bug_get_lines`,
but not like how you did it.

That filetype check will run very frequently, so loading files inside of it is not something I want to see here

folke:
And please change it to `line_length`

chrisands:
From issue
>  It may be slow to find and count the max line length, in which case we could also allow configuring the maximum number of lines to search from the top. Note that we can't always assume the first line will be long (e.g. it could be a short license comment with the long minified lines further down).

@folke is it okay to make configurable how many lines it can scan? 
I would like to make minimum two lines since first line can be short license line and second be the long one.

folke:
I did it a little differently.
Instead of checking max line length, I changed the check to average line length.

Calculating average line length is just `(size - lines)/lines`, so won't impact performance,
and will trigger for minified files.

PR #586, opened 2025-01-17
fix(picker): show all files in git status

kristijanhusak:
## Description

When new folder is created in a git project, and there are files inside, git status picker only shows the folder without any preview. Adding `-uall` to the commands lists them all.

## Screenshots

Before: 
screenshot_2025_01_17_15_42_14

After: 
screenshot_2025_01_17_15_43_13


folke:
ty!

PR #588, opened 2025-01-17
feat(picker): custom icon for unselected entries

Biggybi:
Snacks picker allows setting an icon for selected list entries.
With this PR, we can also select one for unselected entries.

I think it looks great.

image

(Replica of #583 which I closed by mistake, sorry about that.)


folke:
Made two changes:
- added option `show_always` which defaults to `false`. The selected column will only be shown when there is a selection
- added option `unselected`, which defaults to `true`. To use the unselected icon.
- added separate hl group for unselected.

Thank you!

Biggybi:
No problem, thank *you* for your amazing work.

PR #592, opened 2025-01-17
fix(picker.keymaps): keymaps normalization

### LINKED ISSUE #565: bug: (picker) c-j mapping conflict

### Neovim version (nvim -v)

NVIM v0.11.0-dev-1545+ga4f575abd8

### Operating system/version

MacOS 15.3

### Describe the bug

The default mapping `"<c-j>"` for `"list_down` was working initially but not anymore after latest update. It turns out that it conflicts with the following custom Luasnip mapping:

```lua
    {
      "<c-j>",
      function()
        if require("luasnip").jumpable(1) then
          require("luasnip").jump(1)
        end
      end,
      mode = { "i", "s" },
      silent = true,
    },
```

If I disable the the mapping in luasnip, then it does work in Snacks Picker. However, this was not an issue initially and it is not an issue in other similar plugins such as Telescope or Mini.pick.

### Steps To Reproduce

1. Install snacks and enable pick with defaults
2. Set the mapping above in luasnip

### Expected Behavior

It should not conflict with luasnip mapping, as it was not an issue before recent update.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

MurdeRM3L0DY:
- Fixes #565

### DIFF
--- a/lua/snacks/picker/config/init.lua
+++ b/lua/snacks/picker/config/init.lua
@@ -10,10 +10,7 @@ M.alias = {
   oldfiles = "recent",
 }
 
-local key_cache = {} ---@type table<string, string>
-
 --- Fixes keys before merging configs for correctly resolving keymaps.
---- For example: <c-s> -> <C-S>
 ---@param opts? snacks.picker.Config
 function M.fix_keys(opts)
   if not (opts and opts.win) then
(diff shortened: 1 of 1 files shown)


folke:
Added a proper fix. This is really only an issue for `<C-J>`.
It is the same key code as `<NL>`. Have the same work-around in which-key for this.

folke:
Your fix would have broken keymaps.

PR #594, opened 2025-01-17
fix(picker): format: one too many spaces for default icon in results …

Biggybi:
The default icon returned from Snacks utils already has a space for padding.

The fix is to only append a space if there's none.

image


folke:
ty!
I'm using mini.icons and apparently mini does not add the extra space

Biggybi:
Different devs, different standards I guess.
I ought to try this one!

PR #600, opened 2025-01-18
feat(picker): add qflist_all action to send all even when already sel…

ADIX7:
Adds a new action `qflist_all` that send all items to quickfix even when there are already selected ones. This complements the "selected or all" behaviour of `qflist`.


folke:
ty

PR #624, opened 2025-01-18
fix(picker): add OS-specific path separator support

warbacon:
## Description

This update ensures proper syntax highlighting for path entries on Windows systems and correctly applies the `filename_first` formatting option. The changes dynamically detect the appropriate path separator based on the operating system.

### Before Changes  
**`filename_first = false` or `filename_first = true`:**  
Before changes

### After Changes  
**`filename_first = false`:**  
After filename_first = false  

**`filename_first = true`:**  
After filename_first = true


folke:
Path display should always use forward slashes. Just fixed that.

PR #679, opened 2025-01-21
feat(picker.smart): add bufdelete actions from buffers picker

minusfive:
## Description

This enables the ability to delete buffers directly from the `smart` picker (or other combined pickers) by adding a check + warning to the `bufdelete` action (ensuring it only operates on buffer items), and adding the same `dd` and `<c-x>` mappings as the `buffers` picker.

## Related Issue(s)

TBD... searching.

## Screenshots

https://github.com/user-attachments/assets/6c5c4b5e-74ed-4778-9f31-a714f7da1853


folke:
ty!

PR #686, opened 2025-01-21
feat(picker): preview window horizontal scrolling

### LINKED ISSUE #676: feature: Picker preview horizontal scrolling
### Did you check the docs?

I am using a vertical layout on my display and I would like to scroll the preview window horizontally as sometimes the line doesn't fit completely in the preview window.

I would like the ability to configure a keymap to scroll the preview window horizontally 

I considered publishing a PR, similar to the way the preview window is scrolled right now.
I understand we need to add it to win.lua like this:

https://github.com/folke/snacks.nvim/blob/233356e6c9fecf6dfcfddac788580a79d6a3ab6e/lua/snacks/win.lua#L483

and add an action here:

https://github.com/folke/snacks.nvim/blob/main/lua/snacks/picker/actions.lua#L296

However, what is worrying me is this line here:

https://github.com/folke/snacks.nvim/blob/main/lua/snacks/win.lua#L144

because in order to scroll we need to do zl however we send one keycode here... I'm not sure we can send two and what happens if someone remaps the zl keymap?

also, I'm not sure what the sidescrolloff option means and if it related...

Thanks a lot!

### Additional context

Love your work!!

acidclouds:
## Description

This is an initial commit for trying to add horizontal scrolling for the picker preview window.

Known issue: when scrolling horizontally with scope indent, the indent chars persist in the preview window and  don't disappear as they should...

Edit:
I just realized this happens only when we are previewing a buffer that is already open. Only in that case the indents are being rendered... if we are previewing a file that is not currently open the indents don't get rendered and therefore there is no issue. Maybe this is a bug? Should we prevent rendering of the indent guides in the preview window? 

## Related Issue(s)

  - Fixes #676 

## Screenshots

image

### DIFF
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -311,6 +311,14 @@ function M.preview_scroll_up(picker)
   picker.preview.win:scroll(true)
 end
 
+function M.preview_scroll_left(picker)
+  picker.preview.win:hscroll(true)
+end
+
+function M.preview_scroll_right(picker)
+  picker.preview.win:hscroll()
+end
+
 function M.inspect(picker, item)
   Snacks.debug.inspect(item)
 end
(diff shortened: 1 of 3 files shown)


folke:
ty!

PR #701, opened 2025-01-22
feat(picker): page-up/dn preview scroll, help improvements

minusfive:
## Description

This PR:

- Adds `<PageUp>` and `<PageDown>` maps to scroll preview (`n`, `i`).
- Adds `<c-/>` map to toggle help (`n`, `i`), similar to telescope and essentially using the same `?` key already used, but usable from `i` mode.
- Adds descriptions to all `input` window key maps and better organizes them (code + display).
- Adds icons to some keymap descriptions.
- Tweaks default help window column widths and removed separator to maximize doc space (I think highlights + alignment are enough to visually separate them, no need for extra chars).

## Related Issue(s)

N/A

## Screenshots

Screenshot 2025-01-22 at 01 42 17

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -144,109 +144,6 @@ local defaults = {
     tagstack = false, -- save the current position in the tagstack
     reuse_win = false, -- reuse an existing window if the buffer is already open
   },
-  win = {
-    -- input window
-    input = {
-      keys = {
-        ["<Esc>"] = "close",
-        ["<C-c>"] = { "close", mode = "i" },
-        -- to close the picker on ESC instead of going to normal mode,
-        -- add the following keymap to your config
-        -- ["<Esc>"] = { "close", mode = { "n", "i" } },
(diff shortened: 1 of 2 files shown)


minusfive:
Was also thinking, perhaps the help could be a source itself? With a single default key to toggle it (users could add more).

It would remove the space restrictions, potentially displayed under categories (grouped a la wk), and combining all picker win keymaps in one.

minusfive:
Perhaps `keymaps` picker can be enhanced with the ability to pass a `windows` filter? Gonna poke.

minusfive:
Welp, turns out this already works with the `keymaps` picker when set to `local` only. I'll keep playing with a cleanish way to wire it up.

minusfive:
Switched this back to draft as I continue to explore options. I'll likely break up into smaller PRs.

minusfive:
Closing this in favor of #742 and #740 . I'm also exploring adding the option of using which-key for help when available, since I already some ways of checking using `utils.on_module`, etc.

PR #716, opened 2025-01-23
fix(picker.icons): fix typo in Nerd Fonts and display the full category name

noirbizarre:
## Description

Thanks for the icons picker 🙇🏼 

I had one in my dotfiles I was going to contribute, but you have been faster.
So there it is, I added some things I had in mine

This PR:
- fix a typo in the Nerd Fonts source name (its plural)
- add The category full names to be consistent with emojis
 

## Screenshots
image

> [!NOTE]
> Would you be interested in some extra sources I was working on ? At least Math and Box drawing ?

> [!NOTE]
> I would also switch the name the source so Source can be capitalize like the Category and I think having Source > Category after is easier to read as the glyph and the icons are actuelly what people search. But this is personal taste, sp I did not submnitted this change

### DIFF
--- a/lua/snacks/picker/source/icons.lua
+++ b/lua/snacks/picker/source/icons.lua
@@ -1,5 +1,21 @@
 local M = {}
 
+local NERDFONTS_SETS = {
+  cod = "Codicons",
+  dev = "Devicons",
+  fa = "Font Awesome",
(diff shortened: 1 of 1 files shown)


folke:
Sweet, ty!

PR #733, opened 2025-01-24
feat(picker): horizontal list scrolling

acidclouds:
## Description

When working with horizontal layout display, it is sometimes necessary to horizontally scroll the list of results, as the full result does not fit in the list window.
Added an option to horizontally scroll the list window the same way preview window is horizontally scrolled.
Added default keymaps in normal mode only as this is a rare case and usually j/k are used to move across list items in normal mode so this behavior should be natural.
This scrolling can be achieved also by moving the focus to the list window by pressing "/" and then moving the cursor, however I usually just leave insert mode to scroll the results.
## Screenshots

image

image


acidclouds:
Gentle bump...
If you don't like this change, let me know and I'll close the PR

Issue #708, opened 2025-01-23
feature(picker.format): truncate dir


qw457812:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

I'm not a fan of `filename_first`. If the path is too long, or screen is too narrow, the result list may not be readable because the filename is not visible.

Like this (this problem is not only for Java projects):

<img width="1600" alt="Image" src="https://github.com/user-attachments/assets/fb1cb4e2-ea6d-464b-9230-577edf7a1c23" />

### Describe the solution you'd like

I use a monkey-patch to truncate dir, copied from [path_display - telescope](https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/utils.lua#L210). 

And I wish there a built-in format option for something like this:

```lua
  {
    "folke/snacks.nvim",
    optional = true,
    opts = function()
      local strings = require("plenary.strings")
      local truncate, strdisplaywidth = strings.truncate, strings.strdisplaywidth

      -- HACK: shorten & truncate dir
      local filename_orig = Snacks.picker.format.filename
      Snacks.picker.format.filename = function(item, picker)
        local ret = filename_orig(item, picker)

        local dir_trunc_len = vim.api.nvim_win_get_width(picker.list.win.win) - 2
        for _, text in ipairs(ret) do
          if text[2] ~= "SnacksPickerDir" then
            dir_trunc_len = dir_trunc_len - strdisplaywidth(text[1])
            if text[2] == "SnacksPickerFile" then
              break
            end
          end
        end
        if picker.opts.format == "buffer" then
          -- see: https://github.com/folke/snacks.nvim/blob/2568f18c4de0f43b15b0244cd734dcb5af93e53f/lua/snacks/picker/format.lua#L461-L464
          dir_trunc_len = dir_trunc_len - 3 - 1 - 2 - 1
        end

        for _, text in ipairs(ret) do
          if text[2] == "SnacksPickerDir" then
            text[1] = U.path.shorten(text[1]) -- irrelevant to this issue
            -- PERF: holding down j during grep for large projects
            text[1] = truncate(text[1], dir_trunc_len, nil, -1)
            break
          end
        end
        return ret
      end
    end,
  },
```

Screen Shot:

<img width="1600" alt="Image" src="https://github.com/user-attachments/assets/0b54b1ac-ae53-45de-b39c-87384b9dc9cd" />

[...]

PR #743, opened 2025-01-24
feat(picker.format): add `auto` and `align` options for path truncate

qw457812:
## Description

- To prevent wasting width of snacks_picker_list, add the `auto` option to `formatters.file.truncate`.
- To make it more intuitive whether truncate or not, add the `align` option to `formatters.file.truncate`.
- Allow `formatters.file.truncate` to be a function to customize.

## Related Issue(s)

- Work on [#708](https://github.com/folke/snacks.nvim/issues/708).

## Screenshots

auto:

image

align:

image

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -133,7 +133,12 @@ local defaults = {
     },
     file = {
       filename_first = false, -- display filename before the file path
-      truncate = 40, -- truncate the file path to (roughly) this length
+      --- Truncate types:
+      --- * number: truncate the file path to (roughly) this length
+      --- * auto: similar to number, but based on the window width
+      --- * align: align filename to the right border of the window
+      ---@type number|"auto"|"align"|fun(item:snacks.picker.Item, picker:snacks.Picker):string
+      truncate = 40,
       filename_only = false, -- only show the filename
     },
     selected = {
--- a/lua/snacks/picker/format.lua
+++ b/lua/snacks/picker/format.lua
@@ -25,8 +25,8 @@ function M.filename(item, picker)
   if not item.file then
     return ret
   end
+
   local path = Snacks.picker.util.path(item) or item.file
-  path = Snacks.picker.util.truncpath(path, picker.opts.formatters.file.truncate or 40, { cwd = picker:cwd() })
   local name, cat = path, "file"
   if item.buf and vim.api.nvim_buf_is_loaded(item.buf) then
     name = vim.bo[item.buf].filetype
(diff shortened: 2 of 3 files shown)


Seblerr:
As someone often working with long file paths, this would be great. Any news here?

qw457812:
@Seblerr Glad to know you like it. I've been [using](https://github.com/qw457812/dotfiles/blob/fe4a04f1689a7f99dd5472ff7f81310d5c6bdef4/dot_config/nvim/lua/plugins/picker.lua#L317-L367) the `align` option for a long time (before this PR created). But I still not sure how to avoid the hard-coding prefixes part:

https://github.com/folke/snacks.nvim/blob/d6ca18398d4b0625a49bbe9e974c65595459d87c/lua/snacks/picker/format.lua#L73-L73

So I wouldn't be surprised if Folke doesn't accept this PR. I just hope Folke (or someone else) is interested and refactor it...

Diaoul:
Thanks for the documented workaround, I'd love for this to be built-in but I can live with that I guess!

okuuva:
> @Seblerr Glad to know you like it. I've been [using](https://github.com/qw457812/dotfiles/blob/fe4a04f1689a7f99dd5472ff7f81310d5c6bdef4/dot_config/nvim/lua/plugins/picker.lua#L317-L367) the `align` option for a long time (before this PR created). But I still not sure how to avoid the hard-coding prefixes part:
> 
> https://github.com/folke/snacks.nvim/blob/d6ca18398d4b0625a49bbe9e974c65595459d87c/lua/snacks/picker/format.lua#L73-L73
> 
> So I wouldn't be surprised if Folke doesn't accept this PR. I just hope Folke (or someone else) is interested and optimize it...

What's the meaning of the prefix there? It isn't obvious from the code, at least to me. Or, it's obviously affecting the truncation length, but where are those values coming from?

qw457812:
@okuuva Sorry for the delay, I totally missed the notification. I left some explanation about prefix above. You can check it  out :)

folke:
I've been making some changes so that it works  in all pickers and offsets/widths are correctly calculated based on the preceding/following texts.

Since you created a pull request from your main branch, I don't seem to be able to push my changes to it, so I opened a new branch with your commits and my additional changes here https://github.com/folke/snacks.nvim/pull/2294

folke:
## Changes in PR #2294

I've made some modifications to this PR in #2294 since I couldn't push to the original branch. Here are my changes:

### 1. Simplified truncation options
Instead of `auto` and `align`, I've changed the `formatters.file.truncate` option to accept three strategies:
- `"left"`: truncates the beginning of the path (e.g., `…/path/to/file.lua`)
- `"center"`: truncates the middle of the path (default, your original behavior, e.g., `~/pro…/file.lua`)
- `"right"`: truncates the end of the path (e.g., `~/projects/long…`)

### 2. Dynamic width calculation with resolve functions
Instead of hardcoding prefixes for different format types, I've implemented a more flexible approach using resolve functions:
- Added `snacks.picker.format.ctx` type with `picker`, `item`, `offset`, and `max_width`
- Added `resolve` callback support in `snacks.picker.Text` 
- The filename formatter now calculates available width dynamically based on actual window width

This eliminates the hardcoded prefix values (3 for git_status, 7 for buffer, 40 for lsp_symbol) by computing the actual offset at render time.

### 3. Enhanced `truncate()` utility
Updated `M.truncate()` to support left-side truncation with a boolean parameter.

The resolve function approach should be more maintainable and accurate since it calculates the actual available space rather than relying on hardcoded constants.

---

folke:
Just merged the changes. Thank you!

qw457812:
@folke Thank you so much for finishing it. I'm so happy to see you handling it.

Just one small question, will anyone use the `truncate = "right"` option? I can't think of a use case for it.

PR #744, opened 2025-01-24
feat(picker): added new switch_to_grep action

rbhanot4739:
## Description

This PR adds a new `switch_to_grep` action that lets you alternate between your currently opened picker and the `grep` picker. If you invoke the `grep` picker directly, then it would switch to the `files` picker. Here is the sample config on to use it

```lua
sources = {
        smart = {
          win = {
            input = {
              keys = {
                ["<c-k>"] = { "switch_grep_files", desc = "Switch to grep", mode = { "i", "v" } },
              },
            },
          },
        },
        recent = {
          win = {
            input = {
              keys = {
                ["<c-k>"] = { "switch_grep_files", desc = "Switch to grep", mode = { "i", "v" } },
              },
            },
          },
        },
        buffers = {
          win = {
            input = {
              keys = {
                ["<c-k>"] = { "switch_grep_files", desc = "Switch to grep", mode = { "i", "v" } },
              },
            },
          },
        },
        files = {
          win = {
            input = {
              keys = {
                ["<c-k>"] = { "switch_grep_files", desc = "Switch to grep", mode = { "i", "v" } },
              },
            },
          },
        },
        grep = {
          win = {
            input = {
              keys = {
                ["<c-k>"] = { "switch_grep_files", desc = "Switch to grep", mode = { "i", "v" } },
              },
            },
          },
        },
      }
```
PS: This is my first OSS contribution, so apologies if I am not doing something the right way here.

## Screenshots

https://github.com/user-attachments/assets/bbf6a3d9-7b90-411f-a0c1-2bb5c8423c75

### DIFF
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -470,4 +470,42 @@ function M.list_scroll_wheel_up(picker)
   end
 end
 
+M.switch_to_grep = nil
+M.grep_alt_picker = nil
+
+---@param picker snacks.Picker
+function M.switch_to_grep(picker, _)
+  local picker_type = picker.opts.source
+  local allowed_pickers = { "files", "buffers", "recent", "smart", "grep" }
+  if not vim.tbl_contains(allowed_pickers, picker_type) then
+    Snacks.notify.warn("Switching to grep is not supported for `" .. picker_type .. "`", { title = "Snacks Picker" })
+    return
+  end
+
+  if picker_type == "grep" then
+    M.switch_to_grep = false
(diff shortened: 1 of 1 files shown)


rbhanot4739:
@folke any feedback on this pls ? Do you think this `action` makes sense, is it useful in general or are there any  issues with this ?

Since this is actually my first PR i am looking for your thoughts, opinions or any other feedback on this.

folke:
not something that should be in snacks, butyou can of course add it to your own config

rbhanot4739:
Sure, but if you don't mind me asking the reason, does it somehow not fit within the design philosophy of snacks ?

PR #745, opened 2025-01-24
feat(win): use which-key for toggle_help when available

minusfive:
## Description

Leverage `which-key` to show the help window when available.

## Related Issue(s)

N/A

## Screenshots

https://github.com/user-attachments/assets/722048ea-c882-4c0a-b7db-512b136db770

https://github.com/user-attachments/assets/e65f7657-bc83-4de0-9bdd-d9f3dce6c5d9

### DIFF
--- a/lua/snacks/win.lua
+++ b/lua/snacks/win.lua
@@ -353,6 +353,12 @@ end
 
 ---@param opts? {col_width?: number, key_width?: number, win?: snacks.win.Config}
 function M:toggle_help(opts)
+  --- if which-key is available, use it
+  if package.loaded["which-key"] then
+    require("which-key").show({ global = false, delay = 0, defer = false })
+    return
+  end
+
   opts = opts or {}
   local col_width, key_width = opts.col_width or 30, opts.key_width or 10
   for _, win in ipairs(vim.api.nvim_list_wins()) do
@@ -362,6 +368,7 @@ function M:toggle_help(opts)
       return
     end
   end
+
   local ns = vim.api.nvim_create_namespace("snacks.win.help")
   local win = M.new(M.resolve({ style = "help" }, opts.win or {}, {
     show = false,


dpetka2001:
Scrolling the `which-key` window with `<C-d>/<C-u>` doesn't work in `helix` mode, because it scrolls the list window of the picker. If you want to go forward with this, then I'd suggest make it configurable with an option so that users can choose what they want.

minusfive:
@dpetka2001 good catch! I'll dig into why / how it works in normal windows, where it takes over those mappings when they collide.

Providing an option is definitely something I thought about, but would like to discuss it further. I mean, it'd be easy, but `which-key` has already solved many problems the built-in help window / layout still has, and has many more features / options out of the box.

So my thinking was: if a user already "opted in" to using `wk`, then it'd be nice if that was used instead of the built-in by default, and gives them a more consistent experience to boot. Of course these keymap collisions (and any remaining bugs) will need to be ironed out, but I suspect this will be an easier task than re-solving everything `wk` already has with the built-in.

That said, it may be a good idea to expose `wk`'s options regardless, so users can configure this specific instance as granularly as they like, with it inheriting the global `wk` options by default. Will look into it.

dpetka2001:
> So my thinking was: if a user already "opted in" to using wk, then it'd be nice if that was used instead of the built-in by default, and gives them a more consistent experience to boot.

I've seen lots of strange stuff from users. Making it configurable would avoid any future issues. Some people might still prefer the builtin help menu instead of the `which-key` one and they should be able to make the choice even if they have `which-key` installed.

zinobobo418:
Ja

lalvarezt:
with the current implementation calling the toggle_help function a second time doesn't hide it. From what I can see in the you call the show function every time, perhaps a check to see if which_key is already active can solve this

minusfive:
> why not do something like this and leave the user the option to bind it to whatever they want
> 
> 
> 
> ```lua
> 
> return {
> 
>   "folke/snacks.nvim",
> 
> 
> 
>   opts = {
> 
>     picker = {
> 
>       -- default unless overridden, can be placed inside any picker
> 
>       win = {
> 
>         input = {
> 
>           keys = {
> 
>             ["<F12>"] = {
> 
>               function()
> 
>                 require("which-key").show({ global = false })
> 
>               end,
> 
>               desc = "Show which-key",
> 
>               mode = { "n" },
> 
>             },
> 
>           },
> 
>         },
> 
>       },
> 
>     },
> 
>   },
> 
> }
> 
> ```

Yeah, agreed, this is definitely the way to go. Closing.

PR #750, opened 2025-01-25
fix(bufdelete): ctrl-c throw error in `fn.confirm`

phanen:
`vim.fn.confirm` don't return 3 on ctrl-c, but throw an error.
```
E5108: Error executing lua Keyboard interrupt
```


folke:
ty!

PR #757, opened 2025-01-26
fix(filter): correctly use `uv.cwd` when `opts.cwd=true`

### LINKED ISSUE #754: bug(picker): wrong detect cwd with same prefix root

### Neovim version (nvim -v)

0.10.3

### Operating system/version

macOS 15.2

### Describe the bug

wrong detect `cwd` with same prefix root

### Steps To Reproduce
https://github.com/user-attachments/assets/ed09732f-efe8-4a6c-ad91-70a96adb7b4b

1. create project structure like below
```
├── comp
│   └── alpha.js
├── comp-backup
│   └── beta.js
└── components
    └── gamma.js
```
2. open all files to save to recent database
3. close neovim
4. change directory to `comp` (`cd comp`)
5. open recent picker `:lua Snacks.picker.recent()`
6. the wrong items in entry

Image

### Expected Behavior

The recent picker should show only the `alpha.js` entry

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    {
      "folke/snacks.nvim",
      opts = {
        picker = {
          sources = {
            recent = {
              filter = {
                cwd = true,
              },
            },
          },
        },
      },
    },
    -- add any other plugins here
  },
})
```

dpetka2001:
## Description
When `opts.cwd=true` the value of `uv.cwd` should be used to set `self.cwd`.

## Related Issue(s)
Fixes #754

### DIFF
--- a/lua/snacks/picker/core/filter.lua
+++ b/lua/snacks/picker/core/filter.lua
@@ -33,7 +33,7 @@ function M.new(picker)
   self.all = not filter or not (filter.cwd or filter.buf or filter.paths or filter.filter)
   self.paths = {}
   local cwd = filter and filter.cwd
-  self.cwd = type(cwd) == "string" and cwd or opts.cwd or uv.cwd() or "."
+  self.cwd = type(cwd) == "string" and cwd or opts.cwd and uv.cwd() or "."
   self.cwd = vim.fs.normalize(self.cwd --[[@as string]], { _fast = true })
   if not self.all and filter then
     self.buf = filter.buf == true and 0 or filter.buf --[[@as number?]]


folke:
`opts.cwd` should be a `string?`.
But I'll add support for `true` in `config` and will look into that original bug.

folke:
The actual issue was because of `truncpath`. Fixed in https://github.com/folke/snacks.nvim/commit/1069d783347a7a5213240e2e12e485ab57e15bd8

dpetka2001:
I don't think that fixed it. I just pulled and can still see the `LICENCE` file from `~/.config/nvim/lua/plugins` directory, when I should not see it. Also the other items that are in the `cwd` are shown with paths relative to `~/.config/nvim` directory. My fix shows them with no relative path to the roor dir.

folke:
Right, I only fixed the path issue. Will fix the filtering

dpetka2001:
I think it should be like `self.cwd = type(cwd) == "string" and cwd or type(cwd) == "boolean" and uv.cwd() or opts.cwd or "."` no?

folke:
Filter check also fixed now

folke:
Can you check if this fixes it for what you were testing?

dpetka2001:
No, it doesn't. Is what I suggested above `self.cwd = type(cwd) == "string" and cwd or type(cwd) == "boolean" and uv.cwd() or opts.cwd or "."` wrong?

folke:
What is your setup? What options do you set and what do you want to achieve / is wrong?

dpetka2001:
Ok @folke I think this all might as well not have been needed in the end. I reverted to https://github.com/folke/snacks.nvim/commit/3a3e79535bb085815d3add9f678d30b98b5f900f before you made all these changes, because I wanted to check something. 

The problem lies in the LazyVim configuration i think. When you type `<leader>fR` that calls `LazyVim.pick("oldfiles", { filter = { cwd = true }})` then the problem is there. But if I do `:lua Snacks.picker.recent({filter = {cwd = true}})` then it shows correctly the the recent files as I would expect. It seems that somewhere in the `LazyVim.pick` command the `opts` get consumed? Or not get passed correctly? 

Sorry about this, should have done more diligent investigation.

dpetka2001:
Yes, even the LazyVim keymap works as expected now. Thank you :smile:

dpetka2001:
I'm sorry again. But it doesn't. I had changed the mapping to `{ "<leader>fR", function() Snacks.picker.recent({filter = {cwd = true}}) end, desc = "Recent (cwd)" }` and that was working. `LazyVim.pick` still doesn't...

Really sorry about my hastiness...

dpetka2001:
And just for completeness this is my whole picker personal configuration
```lua
picker = {
        win = {
          input = {
            keys = {
              ["<Esc>"] = { "close", mode = { "n", "i" } },
            },
          },
        },
        formatters = {
          file = {
            filename_first = true,
          },
        },
        sources = {
          smart = {
            filter = {
              paths = {
                [vim.fn.stdpath("data")] = false,
                [vim.fn.stdpath("cache")] = false,
                [vim.fn.stdpath("state")] = false,
              },
            },
          },
          projects = {
            win = {
              input = {
                keys = {
                  ["<c-g>"] = { "live_grep", mode = { "i", "n" } },
                },
              },
            },
            actions = {
              live_grep = function(_, item)
                Snacks.picker.pick("grep", {
                  cwd = item.file,
                })
              end,
            },
            confirm = function(_, item)
              Snacks.picker.pick("files", {
                cwd = item.dir,
              })
            end,
          },
          -- Credit https://github.com/folke/snacks.nvim/issues/532#issuecomment-2609303872
          files_with_symbols = {
            multi = { "files", "lsp_symbols" },
            filter = {
              ---@param p snacks.Picker
              ---@param filter snacks.picker.Filter
              transform = function(p, filter)
                local symbol_pattern = filter.pattern:match("^.-@(.*)$")
                local line_nr_pattern = filter.pattern:match("^.-:(%d*)$")
                local search_pattern = filter.pattern:match("^.-#(.*)$")
                local pattern = symbol_pattern or line_nr_pattern or search_pattern

                if pattern then
                  local item = p:current()
                  if item and item.file then
                    filter.meta.buf = vim.fn.bufadd(item.file)
                  end
                end

                if not filter.meta.buf then
                  filter.source_id = 1
                  return
                end

                if symbol_pattern then
                  filter.pattern = symbol_pattern
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 2
                  return
                end

                if line_nr_pattern then
                  filter.pattern = filter.pattern:gsub(":%d*$", "")
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 1
                  local item = p:current()
                  if item then
                    item.pos = { tonumber(line_nr_pattern) or 1, 0 }
                    p.preview:loc()
                  end
                  return
                end

                if search_pattern then
                  filter.pattern = filter.pattern:gsub("#.*$", "")
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 1
                  if search_pattern == "" then
                    return
                  end
                  local item = p:current()
                  vim.api.nvim_buf_call(p.preview.win.buf, function()
                    vim.api.nvim_win_set_cursor(0, { 1, 0 })
                    local search = vim.fn.searchpos(search_pattern, "cw")
                    if search[1] > 0 then
                      vim.cmd("/" .. search_pattern)
                      vim.api.nvim_win_set_cursor(0, { search[1], search[2] })
                      item.pos = { search[1], search[2] }
                    end
                  end)
                  return
                end
              end,
            },
            win = {
              input = {
                keys = {
                  ["<c-n>"] = { "next_result", mode = { "i", "n" } },
                  ["<esc>"] = { "close", mode = { "i", "n" } },
                },
              },
            },
            actions = {
              close = function(p, _)
                vim.cmd("noh")
                p:close()
              end,
              next_result = function(p, _)
                local filter = p:filter()
                local pattern = filter.pattern
                local search_pattern = pattern:match("^.-#(.*)$")

                if search_pattern and search_pattern ~= "" then
                  local item = p:current()

                  vim.api.nvim_buf_call(p.preview.win.buf, function()
                    local search = vim.fn.searchpos(search_pattern, "cw")
                    if search[1] > 0 then
                      vim.cmd("/" .. search_pattern)

                      vim.api.nvim_win_set_cursor(0, { search[1], search[2] })
                      item.pos = { search[1], search[2] }
                    end
                  end)
                  return
                end
              end,
            },
          },
          lsp_symbols = {
            layout = {
              preview = "main",
              preset = "ivy",
            },
          },
        },
      },
```
But I don't see anything that could cause such behavior.

folke:
Right, that's because of root dir being passed as cwd.
So should probably be changed to `Snacks.pick("oldfiles", {cwd = true, ...`

dpetka2001:
Ohhh, I see. That's why when I `dd(filter)` in function `oldfiles` i see the `cwd` is the root dir and `Snacks.picker.recent` prints the correct `cwd`. So, we could just pass `root = false` to `LazyVim.pick`? That should do the same right?

folke:
Probbaly better to use the correct Snacks picker. No need to go through LazyVim.pick for this I guess?

folke:
Feel free to create a PR :)

PR #761, opened 2025-01-26
fix(filter): insert path from `filter.paths` into `self.paths`

dpetka2001:
## Description
Here we see that we insert the paths into `filter` instead of `self.paths` https://github.com/folke/snacks.nvim/blob/b96bd540f785c725289f9f15f0147b1b2dac5a35/lua/snacks/picker/core/filter.lua#L34-L44
 and later on here https://github.com/folke/snacks.nvim/blob/b96bd540f785c725289f9f15f0147b1b2dac5a35/lua/snacks/picker/core/filter.lua#L90-L94 we check `self.paths` to get the path.

This was causing issues to not correctly filter out the already defined paths and I was able to see, for example the scratch buffer files in the pickers.

## Related Issue(s)
None

### DIFF
--- a/lua/snacks/picker/core/filter.lua
+++ b/lua/snacks/picker/core/filter.lua
@@ -40,7 +40,7 @@ function M.new(picker)
     self.buf = self.buf == 0 and vim.api.nvim_get_current_buf() or self.buf
     self.file = self.buf and vim.fs.normalize(vim.api.nvim_buf_get_name(self.buf), { _fast = true }) or nil
     for path, want in pairs(filter.paths or {}) do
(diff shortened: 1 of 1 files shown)


dpetka2001:
Another thing I noticed is that in the `smart` picker one has to redefine in his personal configuration the `filter.paths` from `recent` picker, presumably because there's something weird going on with the merge of the sources configs, since `filter.paths` is a list.

folke:
typo indeed. Ty!

Will see what's wrong with `smart`

folke:
Should be fixed now.
Issue was that the filter was already created, so the source specific filter settings were never applied for multi pickers.

PR #774, opened 2025-01-28
fix(dashboard): don't override user configuration

soulsoiledit:
## Description

When `picker` is enabled, the original line removes it from the table and reinserts it as the first element, preventing the user configuration in `preset.picker` from being used as the picker. This change inserts it as the second element instead.


folke:
Makes sense. ty!

PR #786, opened 2025-01-29
fix(win): error when enabling padding with `listchars=""`

PFiS1737:
## Description

`vim.split` would split `""` into `{ "" }`, causing `table.concat` to output `",eol: "`.

Resolved by using `vim.tbl_filter` to filter out empty strings before applying `table.concat`.


folke:
ty!

PR #789, opened 2025-01-29
fix(lazygit): allow extensible user args

dpetka2001:
## Description
Currently `opts.args` are set to a specific value. Allow users to optionally pass their own additional args.

## Related Issue(s)
None


folke:
ty!

PR #793, opened 2025-01-30
feat(picker): new file explorer `Snacks.picker.explorer()`

folke:
## Description

File explorer picker.

## Testing

```lua
{
  { "nvim-neo-tree/neo-tree.nvim", enabled = false },
  {
    "folke/snacks.nvim",
    keys = {
      { "<leader>e", function() Snacks.picker.explorer() end, desc = "Explorer Snacks" },
    },
    init = function()
      vim.api.nvim_create_autocmd("BufEnter", {
        group = vim.api.nvim_create_augroup("snacks_explorer_start_directory", { clear = true }),
        desc = "Start Snacks Explorer with directory",
        once = true,
        callback = function()
          local dir = vim.fn.argv(0) --[[@as string]]
          if dir ~= "" and vim.fn.isdirectory(dir) == 1 then
            Snacks.picker.explorer({ cwd = dir })
          end
        end,
      })
    end,
  },
}
```

## Screenshots

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,9 @@
 /debug
 /doc/tags
 foo.*
+foo*.*
 node_modules
 tt.*
+/tmp
+repomix*
+.aider*
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -132,7 +132,9 @@ function M.new(opts)
     local sp = vim.fn.screenpos(self.root.win, 1, 1)
     if not vim.deep_equal(sp, self.screenpos) then
       self.screenpos = sp
-      self:update()
+      return self:update()
+    elseif vim.tbl_contains(vim.v.event.windows, self.root.win) then
+      return self:update()
     end
   end)
 
@@ -376,6 +378,15 @@ end
 ---@param parent snacks.win.Dim
 ---@private
 function M:dim_box(widget, parent)
+  -- honor the actual window size for split layouts
+  if not self.opts.fullscreen and widget.id == 1 and self.split and self.root:valid() then
+    return {
+      height = vim.api.nvim_win_get_height(self.root.win),
+      width = vim.api.nvim_win_get_width(self.root.win),
+      col = 0,
+      row = 0,
+    }, { left = 0, right = 0, top = 0, bottom = 0 }
+  end
   local opts = vim.deepcopy(widget) --[[@as snacks.win.Config]]
   -- adjust max width / height
   opts.max_width = math.min(parent.width, opts.max_width or parent.width)
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -5,6 +5,7 @@
 ---@field box_wins snacks.win[]
 ---@field win_opts table<string, snacks.win.Config>
 ---@field closed? boolean
+---@field split? boolean
 ---@field screenpos number[]?
 local M = {}
 M.__index = M
@@ -50,6 +51,7 @@ function M.new(opts)
   -- wrap the split layout in a vertical box
   -- this is needed since a simple split window can't have borders/titles
   if self.opts.layout.position and self.opts.layout.position ~= "float" then
+    self.split = true
     local inner = self.opts.layout
     self.opts.layout = {
       zindex = inner.zindex or 30,
@@ -210,20 +212,23 @@ function M:update()
   end
 
   -- Calculate offsets for vertical splits
-  local voffset = 0
+  local top, bottom = 0, 0
   local pos = self.opts.layout.position
-  if pos and (pos == "left" or pos == "right") then
-    voffset = (vim.o.cmdheight + (vim.o.laststatus == 3 and 1 or 0)) or 0
-    voffset = voffset
-      + (((vim.o.showtabline == 2 or (vim.o.showtabline == 1 and #vim.api.nvim_list_tabpages() > 1)) and 1 or 0) or 0)
+  if pos and (pos == "left" or pos == "right") or self.opts.fullscreen then
+    bottom = (vim.o.cmdheight + (vim.o.laststatus == 3 and 1 or 0)) or 0
+    top = (vim.o.showtabline == 2 or (vim.o.showtabline == 1 and #vim.api.nvim_list_tabpages() > 1)) and 1 or 0
   end
   self:update_box(layout, {
     col = 0,
-    row = 0,
+    row = self.opts.fullscreen and self.split and top or 0, -- only needed for fullscreen splits
     width = vim.o.columns,
-    height = vim.o.lines - voffset,
+    height = vim.o.lines - top - bottom,
   })
 
+  -- fix fullscreen float layouts
+  if self.opts.fullscreen and not self.split then
+    self.root.opts.row = self.root.opts.row + top
+  end
   for _, win in pairs(self:get_wins()) do
     if win:valid() then
       -- update windows with eventignore=all
@@ -405,6 +410,11 @@ function M:update_win(win, parent)
       zindex = self.root.opts.zindex + win.depth,
     }
   )
+  -- fix fullscreen for splits
+  if self.opts.fullscreen and self.split then
+    w.opts.relative = "editor"
+    w.opts.win = nil
+  end
   -- adjust max width / height
   w.opts.max_width = math.min(parent.width, w.opts.max_width or parent.width)
   w.opts.max_height = math.min(parent.height, w.opts.max_height or parent.height)
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -52,10 +52,12 @@ function M.new(opts)
   if self.opts.layout.position and self.opts.layout.position ~= "float" then
     local inner = self.opts.layout
     self.opts.layout = {
+      zindex = inner.zindex or 30,
       box = "vertical",
       position = inner.position,
       width = inner.width,
       height = inner.height,
+      backdrop = inner.backdrop,
       inner,
     }
     inner.width, inner.height, inner.col, inner.row, inner.position = 0, 0, 0, 0, nil
@@ -71,9 +73,18 @@ function M.new(opts)
       local has_border = box.border and box.border ~= "" and box.border ~= "none"
       local is_root = box.id == 1
       if is_root or has_border then
-        local backdrop = false ---@type boolean?
-        if is_root then
-          backdrop = nil
+        local backdrop = box.backdrop
+        if backdrop == nil then
+          backdrop = 60
+        end
+        if is_root and backdrop then
+          backdrop = type(backdrop) == "number" and { blend = backdrop } or backdrop
+          backdrop = backdrop == true and {} or backdrop
+          ---@cast backdrop snacks.win.Backdrop
+          backdrop.win = backdrop.win or {}
+          backdrop.win.zindex = 20
+        else
+          backdrop = false
         end
         self.box_wins[box.id] = Snacks.win(Snacks.win.resolve(box, {
           relative = is_root and (box.relative or "editor") or "win",
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -5,6 +5,7 @@
 ---@field box_wins snacks.win[]
 ---@field win_opts table<string, snacks.win.Config>
 ---@field closed? boolean
+---@field screenpos number[]?
 local M = {}
 M.__index = M
 
@@ -111,6 +112,17 @@ function M.new(opts)
     end
   end)
 
+  self.root:on("WinResized", function(_, ev)
+    if self.closed then
+      return true
+    end
+    local sp = vim.fn.screenpos(self.root.win, 1, 1)
+    if not vim.deep_equal(sp, self.screenpos) then
+      self.screenpos = sp
+      self:update()
+    end
+  end)
+
   -- update layout on VimResized
   self.root:on("VimResized", function()
     self:update()
@@ -183,6 +195,7 @@ function M:update()
   end
   if not self.root:valid() then
     self.root:show()
+    self.screenpos = vim.fn.screenpos(self.root.win, 1, 1)
   end
 
   -- Calculate offsets for vertical splits
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -104,9 +104,21 @@ end
 M.cancel = "close"
 M.edit = M.jump
 M.confirm = M.jump
-M.edit_split = { action = "jump", cmd = "split" }
-M.edit_vsplit = { action = "jump", cmd = "vsplit" }
-M.edit_tab = { action = "jump", cmd = "tabnew" }
+M.edit_split = { "split", "confirm" }
+M.edit_vsplit = { "vsplit", "confirm" }
+M.edit_tab = { "tab", "confirm" }
+
+function M.split()
+  vim.cmd("split")
+end
+
+function M.vsplit()
+  vim.cmd("vsplit")
+end
+
+function M.tab()
+  vim.cmd("tabnew")
+end
 
 function M.toggle_maximize(picker)
   picker.layout:maximize()
@@ -361,6 +373,7 @@ function M.load_session(picker)
 end
 
 function M.help(picker)
+  dd("help")
   local item = picker:current()
   if item then
     picker:close()
@@ -453,6 +466,7 @@ end
 function M.toggle_hidden(picker)
   local opts = picker.opts --[[@as snacks.picker.files.Config]]
   opts.hidden = not opts.hidden
+  picker.list:set_target()
   picker:find()
 end
 
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -448,12 +448,6 @@ function M.focus_preview(picker)
   picker.preview.win:focus()
 end
 
-function M.toggle_ignored(picker)
-  local opts = picker.opts --[[@as snacks.picker.files.Config]]
-  opts.ignored = not opts.ignored
-  picker:find()
-end
-
 function M.item_action(picker, item, action)
   if item.action then
     picker:norm(function()
@@ -461,19 +455,6 @@ function M.item_action(picker, item, action)
       item.action(picker, item, action)
     end)
   end
-end
-
-function M.toggle_hidden(picker)
-  local opts = picker.opts --[[@as snacks.picker.files.Config]]
-  opts.hidden = not opts.hidden
-  picker.list:set_target()
-  picker:find()
-end
-
-function M.toggle_follow(picker)
-  local opts = picker.opts --[[@as snacks.picker.files.Config]]
-  opts.follow = not opts.follow
-  picker:find()
 end
 
 function M.list_top(picker)
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -421,14 +421,10 @@ function M.cycle_win(picker)
   end
   win = wins[idx % #wins + 1] or 1 -- cycle
   vim.api.nvim_set_current_win(win)
-  if win == picker.input.win.win then
-    vim.cmd("startinsert!")
-  end
 end
 
 function M.focus_input(picker)
   picker.input.win:focus()
-  vim.cmd("startinsert!")
 end
 
 function M.focus_list(picker)
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -83,6 +83,7 @@ local M = {}
 ---@field prompt? string prompt text / icon
 ---@field title? string defaults to a capitalized source name
 ---@field auto_close? boolean automatically close the picker when focusing another window (defaults to true)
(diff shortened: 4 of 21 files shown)


dpetka2001:
2 things I noticed around the search function when you press `/`:
- If you have `<esc>` mapped to close the picker, then it leaves you at `insert` mode. Similar thing happened with other pickers, when `stopinsert` was executing after `picker:close()`.
- Since, the search happens in the `cwd`, it would be desirable to search through the folders that are closed in the explorer, but still in the path of `cwd`. Otherwise, you have to open the folder and then do the search, which kind of defeats the purpose of search in my personal opinion.

A separate issue is, when pressing `<C-l>` (which in LazyVim is mapped for `<C-w>l`), I get an error 
```lua
E5108: Error executing lua: /home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:484: Invalid 'window': Expected Lua number
stack traceback:
	[C]: in function 'nvim_win_call'
	/home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:484: in function 'hscroll'
	...ojects/plugins/snacks.nvim/lua/snacks/picker/actions.lua:397: in function 'fn'
	/home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:339: in function </home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:334>
```
Whereas `<C-w>l` works just fine.

dpetka2001:
Also pressing `<CR>` on a folder opens the folder, but when you press it on a file to open the file, it doesn't open it. It just closes the explorer and returns to the already opened buffer.

pkazmier:
> * Since, the search happens in the `cwd`, it would be desirable to search through the folders that are closed in the explorer, but still in the path of `cwd`. Otherwise, you have to open the folder and then do the search, which kind of defeats the purpose of search in my personal opinion.

Yes, I just came here to say the same thing.

dpetka2001:
It seems that `<CR>` for opening a file only works from the input field after you've done a search. But not from the list window when you traverse the explorer tree.

nikbrunner:
I tested the pr branch while it was still open, and also noticed these two things first.

---

> A separate issue is, when pressing `<C-l>` (which in LazyVim is mapped for `<C-w>l`), I get an error
> 
> ```lua
> E5108: Error executing lua: /home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:484: Invalid 'window': Expected Lua number
> stack traceback:
> 	[C]: in function 'nvim_win_call'
> 	/home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:484: in function 'hscroll'
> 	...ojects/plugins/snacks.nvim/lua/snacks/picker/actions.lua:397: in function 'fn'
> 	/home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:339: in function </home/jrn23/projects/plugins/snacks.nvim/lua/snacks/win.lua:334>
> ```
> 
> Whereas `<C-w>l` works just fine.

I think the issue is here, that it conflicts with these default config mappings:

https://github.com/folke/snacks.nvim/blob/927f50200e904d768ae8b711e9dcd0a2accdfd9d/lua/snacks/picker/config/defaults.lua?plain=1#L260-L261

---

> Also pressing `<CR>` on a folder opens the folder, but when you press it on a file to open the file, it doesn't open it. It just closes the explorer and returns to the already opened buffer.

dpetka2001:
Oh, nice catch @nikbrunner. I never used that feature, so wasn't aware of it. I guess it makes sense and can maybe be ignored if Folke doesn't deem it a problem. I also noticed that the corresponding mappings in `defaults.lua` are not the same for `input` and `list`, they are reversed.

dpetka2001:
I also noticed 2 instances of `dd` here https://github.com/folke/snacks.nvim/blob/fa20147abd64b863de62488ed60df448a9c113f2/lua/snacks/picker/actions.lua#L375-L382 and here https://github.com/folke/snacks.nvim/blob/fa20147abd64b863de62488ed60df448a9c113f2/lua/snacks/picker/source/explorer.lua#L245-L248 which you might want to remove :smile:

nikbrunner:
Yeah, I also didn't know about it. I was trying to disable those keymaps in my config, but I couldn't make it work.

When I put in blank strings, I don't get the error, but it also doesn't do anything.
```lua
 ["<c-l>"] = "", 
 ["<c-h>"] = "", 
```

When I use `nil`, the default mapping from snacks gets mapped, and I get an error again.

```lua
 ["<c-l>"] = nil, 
 ["<c-h>"] = nil, 
```

Since the PR is already merged, we should probably create some issues instead of discussing it here? Folke can close it when he doesn't deem it to be an issue. Currently I see three main issues:

- Pressing enter on an entry in the list doesn't open the file. (It does open folders though)
- Keymapping conflict for often used <kbd>Ctrl-h/j/k/l</kbd> navigation bindings
- Error when pressing <kbd>BS</kbd> to navigate up in the tree. (This is also a default binding in the snacks config)

```
E5108: Error executing lua: ...m/lazy/snacks.nvim/lua/snacks/picker/source/explorer.lua:246: attempt to call global 'dd' (a nil value)
stack traceback:
        ...m/lazy/snacks.nvim/lua/snacks/picker/source/explorer.lua:246: in function 'fn'
        ...br/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:339: in function <...br/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:334>
```

dpetka2001:
The horizontal preview mappings have been removed if you check the latest commits, so no prob there. 

`<BS>` correctly navigates up one directory for me without throwing any error. The error you get from pressing `<BS>` is from a remnant of `dd` as also indicated in the error that you pasted, that I've already mentioned in my previous comment and I'm sure this will get removed as well.

Only pressing `<CR>` in the list window remains, which I'm sure it will be solved.

folke:
Just pushed a bunch of changes, so please update :)

- that `<c-l>` was indeed for horizontal scrolling in the preview. I fixed it so it no longer errors when the preview is open and I also removed the two related keymaps. Users can still add them in their own config if they want
- `dd` removed :)
- search: should work as you would expect now. Please test! Since this was a pretty big change of what I had before.

Will check that `<CR>` thing.

pkazmier:
While searching now works for items in closed folders, the cursor line does not highlight the match, so pressing enter does not open the file:

image

dpetka2001:
Also if you would be kind, please look into the minor problem, that when someone has `esc` mapped to close pickers from `insert` mode in the input field when you do a search with `/`, then it leaves you in `insert` mode in the buffer. I believe somewhere a `stopinsert` has to be implemented just like in the other pickers.

Also, with the new live grep it shows correctly the files in the closed folder as well, but I noticed something. Let's say I'm in the buffer `~/.config/nvim/lua/config/keymaps.lua` and search for `snacks` there's a file `~/.config/nvim/lua/plugins/snacks.lua`and when I press `<CR>` on it, then it opens a new empty buffer instead of the correct buffer. I pressed `<C-g>` to show the path of the buffer and I got back 
```lua
"~/.config/LazyVimDev//home/jrn23/.config/LazyVimDev/lua/plugins/snacks.lua" [New] --No lines in buffer--
```
I believe the path is not applied correctly and should be taken into account.

folke:
Found that `<cr>` issue.
Problem was that the `jump` action, sets the current window to the main window where you want to open the file.
The file explorer's file follow code then kicked in and chanched the selected item back to the file already open.
So jump then basically opened the same file :)

folke:
> While searching now works for items in closed folders, the cursor line does not highlight the match, so pressing enter does not open the file. In this screenshot, my `config` folder was closed. So while it does show the contents correctly, the cursor line is not lined up with the matching input.

Right, changed it to always go to the first non-internal node.

folke:
> What is your expectation if you disable live mode and typed: `la` in the fuzzy matcher in your lazyvim config directory? Would you expect it to open the closed folders like `lua` would when live mode is enabled?

This works as expected. When disabling live search, it's just the matcher that's doing its work. Changing the code so that new internal nodes sorted in between other items while matching is not something I'm willing to add. Complex and would be bad for performance.

nikbrunner:
I have updated to the latest version, but I encountered another issue.

When I use the fuzzy search in the explorer and press Enter to select the first matched entry without leaving the input, the file does not open. The path starts from `/Users/..`.

The same issue occurs when I leave the input filled, navigate the list, and then press Enter.

I created a short video for the first case. First, I reproduce the issue, and then I open the same file using the `Snacks.recent` picker to demonstrate the difference in the path shown in the status line.

https://github.com/user-attachments/assets/eea4cc01-60b6-4aeb-9382-9e0da63f01eb

folke:
@dpetka2001 fixed that issue with paths.

Will now look into:

> Also if you would be kind, please look into the minor problem, that when someone has esc mapped to close pickers from insert mode in the input field when you do a search with /, then it leaves you in insert mode in the buffer. I believe somewhere a stopinsert has to be implemented just like in the other pickers.

folke:
@dpetka2001 how can I reproduce this?
I mapped esc to close and when I'm in the input field and do esc it just closes and I'm back in normal mode?

dpetka2001:
I only have this for the relevant part of the config
```lua
opts = {
  picker = {
    win = {
          input = {
            keys = {
              ["<Esc>"] = { "close", mode = { "n", "i" } },
            },
          },
        },    
  },
}
```
and then this 
```lua
{
    "folke/snacks.nvim",
    keys = {
      {
        "<leader>e",
        function()
          Snacks.picker.explorer()
        end,
        desc = "Explorer Snacks",
      },
    },
    init = function()
      vim.api.nvim_create_autocmd("BufEnter", {
        group = vim.api.nvim_create_augroup("snacks_explorer_start_directory", { clear = true }),
        desc = "Start Snacks Explorer with directory",
        once = true,
        callback = function()
          local dir = vim.fn.argv(0) --[[@as string]]
          if dir ~= "" and vim.fn.isdirectory(dir) == 1 then
            Snacks.picker.explorer({ cwd = dir })
          end
        end,
      })
    end,
  },
```
as instructed in the PR. My whole picker config is the following, but I don't think it's relevant
```lua
picker = {
        win = {
          input = {
            keys = {
              ["<Esc>"] = { "close", mode = { "n", "i" } },
            },
          },
        },
        formatters = {
          file = {
            filename_first = true,
          },
        },
        sources = {
          projects = {
            win = {
              input = {
                keys = {
                  ["<c-g>"] = { "live_grep", mode = { "i", "n" } },
                },
              },
            },
            actions = {
              live_grep = function(_, item)
                Snacks.picker.pick("grep", {
                  cwd = item.file,
                })
              end,
            },
            confirm = function(_, item)
              Snacks.picker.pick("files", {
                cwd = item.dir,
              })
            end,
          },
          -- Credit https://github.com/folke/snacks.nvim/issues/532#issuecomment-2609303872
          files_with_symbols = {
            multi = { "files", "lsp_symbols" },
            filter = {
              ---@param p snacks.Picker
              ---@param filter snacks.picker.Filter
              transform = function(p, filter)
                local symbol_pattern = filter.pattern:match("^.-@(.*)$")
                local line_nr_pattern = filter.pattern:match("^.-:(%d*)$")
                local search_pattern = filter.pattern:match("^.-#(.*)$")
                local pattern = symbol_pattern or line_nr_pattern or search_pattern

                if pattern then
                  local item = p:current()
                  if item and item.file then
                    filter.meta.buf = vim.fn.bufadd(item.file)
                  end
                end

                if not filter.meta.buf then
                  filter.source_id = 1
                  return
                end

                if symbol_pattern then
                  filter.pattern = symbol_pattern
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 2
                  return
                end

                if line_nr_pattern then
                  filter.pattern = filter.pattern:gsub(":%d*$", "")
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 1
                  local item = p:current()
                  if item then
                    item.pos = { tonumber(line_nr_pattern) or 1, 0 }
                    p.preview:loc()
                  end
                  return
                end

                if search_pattern then
                  filter.pattern = filter.pattern:gsub("#.*$", "")
                  filter.current_buf = filter.meta.buf
                  filter.source_id = 1
                  if search_pattern == "" then
                    return
                  end
                  local item = p:current()
                  vim.api.nvim_buf_call(p.preview.win.buf, function()
                    vim.api.nvim_win_set_cursor(0, { 1, 0 })
                    local search = vim.fn.searchpos(search_pattern, "cw")
                    if search[1] > 0 then
                      vim.cmd("/" .. search_pattern)
                      vim.api.nvim_win_set_cursor(0, { search[1], search[2] })
                      item.pos = { search[1], search[2] }
                    end
                  end)
                  return
                end
              end,
            },
            win = {
              input = {
                keys = {
                  ["<c-n>"] = { "next_result", mode = { "i", "n" } },
                  ["<esc>"] = { "close", mode = { "i", "n" } },
                },
              },
            },
            actions = {
              close = function(p, _)
                vim.cmd("noh")
                p:close()
              end,
              next_result = function(p, _)
                local filter = p:filter()
                local pattern = filter.pattern
                local search_pattern = pattern:match("^.-#(.*)$")

                if search_pattern and search_pattern ~= "" then
                  local item = p:current()

                  vim.api.nvim_buf_call(p.preview.win.buf, function()
                    local search = vim.fn.searchpos(search_pattern, "cw")
                    if search[1] > 0 then
                      vim.cmd("/" .. search_pattern)

                      vim.api.nvim_win_set_cursor(0, { search[1], search[2] })
                      item.pos = { search[1], search[2] }
                    end
                  end)
                  return
                end
              end,
            },
          },
          lsp_symbols = {
            layout = {
              preview = "main",
              preset = "ivy",
            },
          },
        },
      },
```
except the part for `win.input.keys`.

folke:
@nikbrunner yep, fixed. Was the same issue reported by @dpetka2001

folke:
@dpetka2001 this part might cause issues:

```lua
close = function(p, _)
                vim.cmd("noh")
                p:close()
              end,
```
You're probably unaware but this will override the `Snacks.win` close action (which is different from picker actions that get wrapped into win actions)

folke:
However, even with that I still can't reproduce it.

folke:
Just tried it with a minimal repro and still can't reproduce.

folke:
Are you using a winbar plugin by any chance?

dpetka2001:
No winbar. Just tried a fresh LazyVim installation and it works there. Will dig in further to find out what's causing this, so please disregard for the time being until I manage to come up with new info or not.

nikbrunner:
A question: Would it be possible that the Explorer Sidebar stays open after I select an item? I personally would like that. But I also can live without it. I have found the `auto_close` config option, which firstly I assumed was for this usecase, but the annotation told me, its for something else:

```lua
---@field auto_close? boolean automatically close the picker when focusing another window (defaults to true)
```

folke:
Just added `opts.jump.close` which defaults to `true` for all, but not for the explorer.
Makes more sense to keep it open indeed.

nikbrunner:
Thats awesome. Thank you 🙏

dpetka2001:
I found the culprit for being dropped into `insert` mode. It's the plugin `barbar.nvim` that I replaced `bufferline.nvim` with. Now need to find out the reason it happens. I don't know if that's what you meant by winbar plugin, but I thought you meant for example a plugin like `dropbar` that adds breadcrumbs in the winbar, that's why I said I didn't have any. And I only use `barbar` to show me the buffers just like bufferline did.

folke:
That's odd indeed! No idea what that plugin does or why it would cause issues.
Prompt buffers however really behave weird and different from normal buffers with regards to stopinsert.
I'm still doing a `stopinsert`, but that's actually not needed because Neovim does that automatically when you leave a prompt buffer.

I added the extra `stopinserts` to fix weird issues with cursor jumps being off by one mostly. and for some users that get put in INSERT mode.

Let me know if you do find anhything!

JensPauwels:
Hey @folke,

Noticed an error in the explorer.

Steps to reproduce:

1) Open explorer
2) Close file window
3) Open file
4) Open explorer again

https://github.com/user-attachments/assets/040b10ba-d65c-40c1-9bb3-5cd933722f70

nikbrunner:
> Just added `opts.jump.close` which defaults to `true` for all, but not for the explorer. Makes more sense to keep it open indeed.

This works now, but now the chosen file doesnt open when picked from the list. It does open when searched for in the input.

https://github.com/user-attachments/assets/24f52694-93a5-4176-aeff-7c0ceff4c43c

I also should find a way to disable the blink completion when inside that explorer input.. 🤔

JensPauwels:
> > Just added `opts.jump.close` which defaults to `true` for all, but not for the explorer. Makes more sense to keep it open indeed.
> 
> This works now, but now the chosen file doesnt open when picked from the list. It does open when searched for in the input.
> 
>  Screen.Recording.2025-01-30.at.22.03.09.mp4 
> I also should find a way to disable the blink completion when inside that explorer input.. 🤔

Indeed, would be nice to have an easy way to disable the blink completion. Had the same feeling

folke:
blink is disabled by default in prompt buffers and the snacks picker input is a prompt buffer, so if that's not working for you it means you probably changed your blink config.

nikbrunner:
I did it like this. working fine for me now: https://github.com/nikbrunner/nbr.nvim/blob/244ec9317429a8efc253c2603324abc5a606a501/lua/nbr/specs/blink.lua?plain=1#L9-L11

folke:
@nikbrunner like I said blink excludes prompt buffers by default, so you can just not confiugure that setting.
See here https://cmp.saghen.dev/configuration/general.html

folke:
@dpetka2001 does it also work if you change it to just `picker:close()`?

JensPauwels:
Is there a way to disable opening a previous file in the tree explorer? (using <C-o ->)

https://github.com/user-attachments/assets/e8baacbf-598a-45c6-b894-d1f67cf6c365

dpetka2001:
Yes, it does. I just did it like this because of what you mentioned in your previous comment about overwriting the default
```lua
close = function(p, _)
                vim.cmd("noh")
                p:close()
              end,
```

> You're probably unaware but this will override the Snacks.win close action (which is different from picker actions that get wrapped into win actions)

fdschmidt93:
As someone who's also done something similar (in fact, sole user of [tree-branch in telescope-file-browser.nvim](https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/210) for 2 years :) and is in the process of switching to Snacks I thought I might leave a few suggestions for additional actions/slight behavior changes. 

1. When being in the prompt, finding a directory, hitting enter could clear the prompt, open the directory node, and then jumping to the directory node in list window

I tried something locally like this 

```lua
  confirm = function(picker)
    local state = M.get_state(picker)
    local item = picker:current()
    if not item then
      return
    elseif item.dir then
      state:toggle(item)
      if vim.api.nvim_get_current_win() == picker.input.win.win then
        picker.input:set("", "")
        vim.defer_fn(function()
          for idx, item_ in ipairs(picker.list.items) do
            if item_.file == item.file then
              picker.list:move(idx)
              break
            end
          end
        end, 100)
      end
    else
      picker:action("jump")
    end
  end,
```
though couldn't quite get it to work. My telescope expertise doesn't translate 1:1 to the Snacks picker yet I'm afraid ;) but maybe that idea is welcome as a default.

2. Copy / yank / move

The way I did this in my file-browser tree branch above is that a user

- Select files/dirs
- Searches in prompt for directory (or for a file in a directory) to move or copy selections to (i.e., executing action. From a quick glance that is not yet supported. I really liked this because it was _fast_.

That said, in my file browser the user practically speaking only used the prompt. So it's quite a different UX.

3. Incremental (recursively) unrolling / collapsing of (sub-)trees
(I realize entering "." in prompt unrolls all trees, but this is more a QoL idea)
When hovering a directory, ">" opens all directories within a directory, "<" does the reverse. Could also be global. Helps with quick targeted directory tree exploration.

4. Batch file renaming of selections in a separate modifiable buffer. Though I fully understand if that would be out of scope.

folke:
@fdschmidt93 you can select multiple files with `<Tab>` or first visual select them and press `<Tab>`.
Then go to a dir and press `m` to move them there.

Good idea on that open of dirs from search. Will add

folke:
@fdschmidt93 added in https://github.com/folke/snacks.nvim/commit/605f7451984f0011635423571ad83ab74f342ed8

folke:
@dpetka2001 added that explicit close function, so your work-around should no longer be needed.
All very weird.
Has probably something to do with `WinClosed` events not being triggered recursively.
I added some work-arounds for that in the win class, but I can't account for all situations, especially not the weird ones from prompt buffers.

UtsavBalar1231:
@folke can you expose an optional configuration for just showing the basename of the directory thats currently visiable on cwd.

screenshot_2025-01-31_11-35-15

This takes away way too much space, just the basename of directory should be enough. and maybe an explorer_help method to get some additional information about the cwd?.

If possible inserting a `../` on top just like other file_explorers would be nice.

folke:
@UtsavBalar1231 that's because you disabled the tree.

moniquelive:
@folke is it possible to sort by extension? The sorting documentation is lacking.

TIA.

Issue #751, opened 2025-01-25
feature: Git files should support viewing everything as if you were at the root of the repo


WolfeCub:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

In telescope and fzf-lua `git_files` would show you all the files from the root of the repo. However in snacks it's relative to your current working directory.

### Describe the solution you'd like

Add another function, option or change the behaviour of `git_files` so that there's a way to view the directory contents as if you were at the root from any sub-directory. The files should be formatted relative to the repo root regardless of where in the repo you are.

### Describe alternatives you've considered

I would write it myself but the documentation around `cwd` and `transform` to get the correct formatting is unclear.

PR #804, opened 2025-01-30
fix(git): fix git root detection

lttb:
## Description

Regression was introduced in 3cdebee880742970df1a1f685be4802b90642c7d (#751)

Fix git root detection for bare repositories.

With bare repositories, the `.git` search in parent folders might not work, in which case the picker fails with:

```
E5108: Error executing lua: vim/shared.lua:0: s: expected string, got nil
stack traceback:
	[C]: in function 'error'
	vim/shared.lua: in function 'validate'
	vim/shared.lua: in function 'startswith'
	vim/fs.lua: in function 'normalize'
	.../nvim/lazy/snacks.nvim/lua/snacks/picker/core/filter.lua:57: in function 'set_cwd'
	.../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:700: in function 'set_cwd'
	...e/nvim/lazy/snacks.nvim/lua/snacks/picker/source/git.lua:28: in function '_find'
	.../nvim/lazy/snacks.nvim/lua/snacks/picker/core/finder.lua:119: in function 'run'
	.../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:741: in function 'find'
	.../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:201: in function 'git_files'
	/Users/lttb/.config/nvim/lua/lttb/plugins/snacks.lua:96: in function </Users/lttb/.config/nvim/lua/lttb/plugins/snacks.lua:95>
```

- this fix introduces `git rev-parse --show-toplevel`, with the previous '.git' search logic as a fallback.
- it also updates the caching logic a bit, to populate the cache with `git_root` on the way to the root directory, to optimise a bit git root detection from other sub directories and avoid system calls

### DIFF
--- a/lua/snacks/git.lua
+++ b/lua/snacks/git.lua
@@ -14,16 +14,21 @@ Snacks.config.style("blame_line", {
   ft = "git",
 })
 
-local git_cache = {} ---@type table<string, string>
+-- `string` for cache hit, `false` for cache miss
+local git_cache = {} ---@type table<string, string | false>
 local function is_git_root(dir)
-  if git_cache[dir] == nil and (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil then
-    git_cache[dir] = vim.fs.normalize(dir)
+  if git_cache[dir] == nil then
+    if (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil then
+      git_cache[dir] = vim.fs.normalize(dir)
+    else
+      git_cache[dir] = false
+    end
   end
   return git_cache[dir]
 end
 local function get_git_root()
-  local git_root = vim.system({ "git", "rev-parse", "--show-toplevel" }, { text = true }):wait().stdout
-  return git_root and vim.fs.normalize(git_root):gsub("[\n\r]", "")
+  local git_root = vim.fn.trim(vim.fn.system("git rev-parse --show-toplevel"))
+  return vim.fs.normalize(git_root)
 end
 
 --- Gets the git root for a buffer or path.
@@ -38,7 +43,7 @@ function M.get_root(path)
   -- check cache first
   for dir in vim.fs.parents(path) do
     if git_cache[dir] then
-      return git_cache[dir]
+      return git_cache[dir] or nil
     end
   end
 
--- a/lua/snacks/git.lua
+++ b/lua/snacks/git.lua
@@ -14,12 +14,16 @@ Snacks.config.style("blame_line", {
   ft = "git",
 })
 
-local git_cache = {} ---@type table<string, boolean>
+local git_cache = {} ---@type table<string, string>
 local function is_git_root(dir)
-  if git_cache[dir] == nil then
-    git_cache[dir] = (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil
+  if git_cache[dir] == nil and (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil then
+    git_cache[dir] = vim.fs.normalize(dir)
   end
   return git_cache[dir]
+end
+local function get_git_root()
+  local git_root = vim.system({ "git", "rev-parse", "--show-toplevel" }, { text = true }):wait().stdout
+  return git_root and vim.fs.normalize(git_root):gsub("[\n\r]", "")
 end
 
 --- Gets the git root for a buffer or path.
@@ -34,11 +38,20 @@ function M.get_root(path)
   -- check cache first
   for dir in vim.fs.parents(path) do
     if git_cache[dir] then
-      return vim.fs.normalize(dir) or nil
+      return git_cache[dir]
     end
   end
+
+  local git_root = get_git_root()
+
   for dir in vim.fs.parents(path) do
-    if is_git_root(dir) then
+    if git_root then
+      git_cache[dir] = git_root
+
+      if vim.fs.normalize(dir) == git_root then
+        return git_root
+      end
+    elseif is_git_root(dir) then
       return vim.fs.normalize(dir) or nil
     end
   end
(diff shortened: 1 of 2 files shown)


lua/snacks/git.lua
```diff
@@ -14,13 +14,17 @@ Snacks.config.style("blame_line", {
...
+local git_cache = {} ---@type table<string, string>
 local function is_git_root(dir)
-  if git_cache[dir] == nil then
-    git_cache[dir] = (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil
+  if git_cache[dir] == nil and (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil then
+    git_cache[dir] = vim.fs.normalize(dir)
   end
   return git_cache[dir]
 end
+local function get_git_root()
```
folke, line 24: vim.system is not available on 0.9, so we can't use that.

lua/snacks/git.lua
```diff
@@ -14,13 +14,17 @@ Snacks.config.style("blame_line", {
   ft = "git",
 })
 
-local git_cache = {} ---@type table<string, boolean>
+local git_cache = {} ---@type table<string, string>
 local function is_git_root(dir)
-  if git_cache[dir] == nil then
-    git_cache[dir] = (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil
+  if git_cache[dir] == nil and (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil then
```
folke, line 19: we also want to cache misses, not only hits

lua/snacks/git.lua
```diff
@@ -34,11 +38,20 @@ function M.get_root(path)
   -- check cache first
   for dir in vim.fs.parents(path) do
     if git_cache[dir] then
-      return vim.fs.normalize(dir) or nil
+      return git_cache[dir]
     end
   end
+
+  local git_root = get_git_root()
```
folke, line 45: why do you always execute `git` here?

lttb: it'll be executed only if there are no parent folders with the cached git directory in the cache, which should probably happen only once as the lookup above will get the cached dir next time

folke:
Can you show me the steps that actually show the issue you're having?

folke:
Are you using GIT_DIR and GIT_WORK_TREE env vars?

lua/snacks/git.lua
lttb: replaced with `vim.fn.system("git rev-parse --show-toplevel")`, which should be supported by 0.9 iirc?

folke:
Just pushed some changes. Can you check if this fixes it?

lttb:
> Can you show me the steps that actually show the issue you're having?
> Are you using GIT_DIR and GIT_WORK_TREE env vars?

yeah I faced this issue with my dotfiles repository (https://github.com/lttb/dotfiles), managed by [yadm](https://github.com/yadm-dev/yadm)

this is the alias I use to edit dotfiles:

```
alias d='\
    DOTFILES=1 \
    GIT_DIR=$HOME/.local/share/yadm/repo.git \
    GIT_WORK_TREE=$HOME \
    nvim -c ":cd ~"'
```

folke:
Then I assume my latest changes fix this?

lttb:
> Just pushed some changes. Can you check if this fixes it?
> Then I assume my latest changes fix this?

thanks for the fix! can confirm it works in my case.

I wonder if `GIT_WORK_TREE` is more robust solution that checking it with git directly?

I can't think of all of the cases, but I think it's pretty standard to use `git `, eg
- in fzf.lua - https://github.com/ibhagwan/fzf-lua/blob/8af88e542d70ed397f297a81168be0dd3394b0d4/lua/fzf-lua/path.lua#L545
- telescope - https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/builtin/__git.lua#L483

also I believe my solution should be slightly more efficient, as it'll avoid unnessary lookups to the top parent for further `get_root` if the `git_root` is detected, which should be the case most of the times

folke:
git rev-parse will do pretty mcuh the same.
Look for a .git repo and fallback to GIT_WORK_TREE if set.
So there;s no need to use rev-parse here. closing

folke:
I'm storing cache misses, so the fs checks are already very minimal.
And calling `git` which will internally also do all those fs checks is way more expensive anyway.

lttb:
> I'm storing cache misses, so the fs checks are already very minimal.
> And calling git which will internally also do all those fs checks is way more expensive anyway.

@folke not sure that's the case with the latest fix though, as `GIT_WORK_TREE` fallback isn't stored anywhere in the cache? so the next `get_root` call will do these checks again, if I'm understanding correctly?

let's consider this case:
- root is `/a/b`, with `GIT_WORK_TREE`
- we call `get_root('/a/b/c/d/e')`
- it'll get to the top and return `env.GIT_WORK_TREE` as a fallback
- the next call `get_root('/a/b/c/d/e/f/c')` - we will do the lookup again, as `GIT_WORK_TREE` isn't cached

so essentially, with git work trees, every `get_root` call will cause a file system lookup.

what do you think about adding `vim.env.GIT_WORK_TREE` check to the cache hit too?

```lua
local function is_git_root(dir)
  if git_cache[dir] == nil then
    git_cache[dir] = (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil
        or vim.fs.normalize(dir) == vim.env.GIT_WORK_TREE
  end
  return git_cache[dir]
end

function M.get_root(path)
  path = path or 0
  path = type(path) == "number" and vim.api.nvim_buf_get_name(path) or path --[[@as string]]
  path = vim.fs.normalize(path)
  path = path == "" and (vim.uv or vim.loop).cwd() .. "/foo" or path
  -- check cache first
    if git_cache[dir] then
      return vim.fs.normalize(dir) or nil
    end
  end
    if is_git_root(dir) then
      return vim.fs.normalize(dir) or nil
    end
  end
end
```

PR #807, opened 2025-01-31
feat(git): optimise git root lookup

lttb:
## Description

A follow-up to #804

This PR contains two improvements for `git.get_root`:

### `path` parents fix

```diff
- path = path == "" and (vim.uv or vim.loop).cwd() .. "/foo" or path
+ path = (path == "" and (vim.uv or vim.loop).cwd() or path) .. "/foo"
  -- check cache first
  for dir in vim.fs.parents(path) do
```

The issue was that in case if `path` is actually a `git_root`, `vim.fs.parents(path)` will go up, so `dir` will never hit the cache. I suppose that was the reason for the `/foo` postfix for `(vim.uv or vim.loop).cwd()` too.

See the issue and the fix combined in the video:

https://github.com/user-attachments/assets/bba89ffa-13b3-40cc-8b44-06826afe68be

### cache hit for `GIT_WORK_TREE`

`GIT_WORK_TREE` is never cached, so any call to `git_root` will cause a full lookup in the file system before returning the `GIT_WORK_TREE`. 

To optimise it, I think it makes sense to add `GIT_WORK_TREE` to `is_git_root` for caching.

See the issue and the fix combined in the video:

https://github.com/user-attachments/assets/f37c960f-2ed3-49cb-9d40-73df64d82bea

## Related Issue(s)

- #804

### DIFF
--- a/lua/snacks/git.lua
+++ b/lua/snacks/git.lua
@@ -15,9 +15,9 @@ Snacks.config.style("blame_line", {
 })
 
 local git_cache = {} ---@type table<string, boolean>
-local function is_git_root(dir)
+local function is_git_root(dir, git_work_tree)
   if git_cache[dir] == nil then
-    git_cache[dir] = (vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil
+    git_cache[dir] = ((vim.uv or vim.loop).fs_stat(dir .. "/.git") ~= nil) or (vim.fs.normalize(dir) == git_work_tree)
   end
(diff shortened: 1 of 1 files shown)


folke:
Checking the top-level was indeed something that had to be added.
Adding GIT_WORK_TREE in the cache is unnecessary and also incorrect imo.

It also doesn't matter since we're caching misses as well.

So now for every `get_root` only the first time will we need to check the roots.
After that it's purely the cache.

PR #823, opened 2025-01-31
feat(scope): allow injected languages to be parsed by treesitter

### LINKED ISSUE #430: feature(scope): Add support for injections in treesitter scopes
### Did you check the docs?

I'm regularly working with .vue files and have noticed that the treesitter scopes do not correctly handle the injected JS/TS/CSS blocks, and simply treat the entire enclosing tag as the scope.

Example:

https://github.com/user-attachments/assets/79506eff-acf4-4dec-a7a3-a0b300169ada

This is due to the calls to `vim.treesitter.get_node()` in `scope.lua` which pass the default argument for `ignore_injections` which is `true`.
With `ignore_injections` set to `false`, we get the desired behaviour:

https://github.com/user-attachments/assets/31764317-dc74-47f7-bf1f-31d51152cbd9

However, I imagine not all users may want injection parsing enabled, so I recommend adding a new config flag that defaults to the current behaviour, but can be toggled to support including the injections if desired.

Example config:
```
scope = {
  enabled = true,
  treesitter = { include_injections = true },
},
```

Perhaps this default is settable from the treesitter side, and doesn't require modification here, but I couldn't find it if so.

### Additional context

I have a patch on my fork that implements this as described above, and would be happy to open a PR if desired.

Aietes:
## Description

Adding a config flag to allow injected languages to be parsed by treesitter. This is crucial for languages like vue, which contain typescript or javascript in a script tag. When `treesitter.get_node` is not called with ignore_injections = false, the script tag is treated as a single block, which is not helpful, as it contains most code for languages like vue.

## Related Issue(s)

 - Fixes #430 

## Screenshots

 - See #430

### DIFF
--- a/lua/snacks/scope.lua
+++ b/lua/snacks/scope.lua
@@ -43,6 +43,7 @@ local defaults = {
     -- detect scope based on treesitter.
     -- falls back to indent based detection if not available
     enabled = true,
+    ignore_injections = false, -- do not ignore injected languages, helpful for languages like vue
     ---@type string[]|{enabled?:boolean}
     blocks = {
       enabled = false, -- enable to use the following blocks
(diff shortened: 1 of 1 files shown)


Aietes:
Opened #906 to ensure PR is considered for merge.

folke:
ty!

PR #872, opened 2025-02-02
feat(picker.sort): index biased score sorting

ADIX7:
## Description

This PR adds an index biased sorting for pickers beside the default and the index based. It is based on [Telescope's](https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/sorters.lua#L412) algorithm

## Related Issue(s)

Not aware

### DIFF
--- a/lua/snacks/picker/sort.lua
+++ b/lua/snacks/picker/sort.lua
@@ -56,4 +56,25 @@ function M.idx()
   end
 end
 
+local index_biased_score = function(score, idx)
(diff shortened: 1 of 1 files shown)


folke:
Just added support for fzf's history scheme.

Set `opts.matcher.history_bonus = true` and tie-break to index with `sort = { fields = { "score:desc", "idx" } }`

PR #879, opened 2025-02-02
feat(picker.commands): do not autorun commands that require arguments

antonk52:
## Description

Currently when you use picker.commands to select a command that requires an argument, it will cause an error. `:BlickCmd` is one example. To avoid the error we can feed the keys to enter the command mode and populate the command and leave it to the user what they want to do. 

This is also what telescope [does](https://github.com/nvim-telescope/telescope.nvim/blob/415af52339215926d705cccc08145f3782c4d132/lua/telescope/builtin/__internal.lua#L399) in its builtin command picker.

nargs values `*` or `?` mean that the command can be executed both with arguments and without. I think it is safer to leave it to the user to decide if they want them to either trigger a run or provide the arguments. But happy to change it if you think running the commands by default makes sense.

### DIFF
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -386,6 +386,13 @@ function M.cmd(picker, item)
   picker:close()
   if item and item.cmd then
     vim.schedule(function()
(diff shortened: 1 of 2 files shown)


folke:
Simplified the code a bit and merged. Thank you!

PR #882, opened 2025-02-03
feat(picker.matcher): add option to enable/disable boundary bonus

linusboehm:
## Description

I have a picker to search my bash history to execute a bash command in a Snacks.terminal. I'd like to have the same order of search results as when using `fzf` with `--scheme=history`.
Using `  sort = { fields = { "score:desc", "idx" } },` together with the new option to disable the boundary bonus, seems to give me the same results.


folke:
Just added a new `history_bonus` option that does exactly the same as fzf's scheme=history

PR #885, opened 2025-02-03
fix(picker.git_log): add extra space between the date and the message

hankertrix:
## Description

Add additional space between the date of the commit and the commit message so that the message is more readable. Also added support for spaces between the type and the scope of the commit.

## Screenshots

Before:

image

image

After:

image

image


folke:
Makes sense. ty!

hankertrix:
You're welcome! Thanks for merging it so quickly!

PR #909, opened 2025-02-03
feat(picker): add create and delete branch to git_branches

IanLiuTW:
## Description

This PR adds two actions: `git_create_branch` and `git_delete_branch`. They are aimed to be included in the `git_branches` picker to easily create/delete git branches with custom keymaps.


folke:
I refactored the code a bit an merged. ty!

PR #911, opened 2025-02-04
fix(picker.actions): fix qflist position

jdrouhard:
vim.fn.setqflist() and vim.fn.setloclist() use (1,1)-indexing, but vim.api.nvim_win_set_cursor() uses (1,0)-indexing. Adjust for this when sending items to qflist/loclist.


folke:
ty!

PR #936, opened 2025-02-05
Set priority on the bigfile pattern so it can be overriden by users

xentac:
For users who have big files of a particular filetype, they can create their own vim.filetype that has a higher priority than bigfile.

## Description

I have a certain filetype that I know is big and I'm willing to accept the speed hit for the features that I get (particularly this is a beancount file, but it shouldn't matter). If I want to override bigfile for these files, I'd like to be able to define a filetype that takes precedence over the bigfile pattern. vim.filetypes.add lets you set a priority, but it will only override filetypes that also have a priority.

My local config after this change looks something like this (lua/plugins/treesitter.lua):

```
return {
  {
    "nvim-treesitter/nvim-treesitter",
    opts = {
      ensure_installed = {
        "beancount",
      },
      function(_, _)
        vim.filetype.add({
          extension = {
            beancount = { "beancount", { priority = 100 } },
          },
        })
      end,
    },
  },
}
```


folke:
In this case any filetype with a priority will never get bigfile?

xentac:
I suppose that is true. I'm open to other suggestions for other ways to do this. The other one I could think of was an exclusion list of regexps to match filenames against.

PR #943, opened 2025-02-05
[dashboard] update docs to add info about `gh notify` extension

ahmedelgabri:
`gh notify` is not a `gh` command, it requires installing the `gh-notify` extension https://github.com/meiji163/gh-notify


folke:
ty

PR #951, opened 2025-02-05
feat(picker): add confirmation before deleting a git branch

IanLiuTW:
## Description

A minor update to add a confirmation dialogue before deleting a branch in `git_branches`.


folke:
Can you change it to use select?

```lua
  Snacks.picker.select({ "Yes", "No" }, { prompt = ... }, function(_, idx)
    if idx == 1 then
      ...
    end
  end)
```

folke:
ty!

PR #970, opened 2025-02-06
feat(picker): do not save empty patterns as last

### LINKED ISSUE #1130: feature(picker):  do not save empty patterns as last
### Did you check the docs?

It occurs sometimes that I launch picker, and I immediately realize that I should have just resumed the last search. However, at this point I cannot do so, because my freshly open (empty) picker will overwrite the last search, and "resuming" at this point would just open an empty picker again.
I think it'd make sense to save last state for resumption only when the currently open picker is not empty. Otherwise, resuming does not really give any benefit.

Do not save picker's last state when the input is empty.

Remembering to use "Resume" feature directly :P

### Additional context

_No response_

marcinjahn:
## Description

Maybe it's just me, but it occurs sometimes that I launch picker, and I immediately realize that I should have just resumed the last search. However, at this point I cannot do so, because my freshly open (empty) picker will overwrite the `M.last`. I think it'd make sense to overwrite `M.last` only when the currently open picker is not empty. Otherwise, resuming does not really give any benefit.

## Related Issues

Closes #1130

### DIFF
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -627,14 +627,16 @@ function M:close()
   self:hist_record(true)
   self.closed = true
 
-  M.last = {
-    opts = self.init_opts or {},
-    selected = self:selected({ fallback = false }),
-    cursor = self.list.cursor,
-    topline = self.list.top,
-    filter = self.input.filter,
-  }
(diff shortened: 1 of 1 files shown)


folke:
this doesnt make sense to me, since not every picker needs a filter.
You can also use search history to go back. closing

PR #991, opened 2025-02-07
feat(lazydocker): added a Lazydocker extension similar to Lazygit

umutondersu:
## Description

  I was using lazydocker with a custom window with the below configuration.

```lua
            Snacks.win({
                style = 'lazygit',
                file = nil,
                buf = vim.api.nvim_create_buf(false, true), -- Create a scratch buffer for the terminal
                on_win = function(self)
                    vim.fn.termopen("lazydocker", {
                        on_exit = function() -- Close the window when the process exits
                            self:close()
                        end,
                    })
                    vim.cmd("startinsert") -- Enter insert mode for terminal
                end,
            })
```
So i thought it would be useful to have a lazydocker QoL plugin like lazygit

## Screenshots

lazydocker

### DIFF
--- a/lua/snacks/lazydocker.lua
+++ b/lua/snacks/lazydocker.lua
@@ -1,0 +1,212 @@
+---@class snacks.lazydocker
+---@overload fun(opts?: snacks.lazydocker.Config): snacks.win
+local M = setmetatable({}, {
+  __call = function(t, ...)
+    return t.open(...)
+  end,
+})
+
+M.meta = {
+  desc = "Open LazyDocker in a float, auto-configure colorscheme and integration with Neovim",
+}
+
+---@alias snacks.lazydocker.Color {fg?:string, bg?:string, bold?:boolean}
+
+---@class snacks.lazydocker.Theme: table<number, snacks.lazydocker.Color>
+---@field activeBorderColor snacks.lazydocker.Color
+---@field cherryPickedCommitBgColor snacks.lazydocker.Color
+---@field cherryPickedCommitFgColor snacks.lazydocker.Color
+---@field defaultFgColor snacks.lazydocker.Color
+---@field inactiveBorderColor snacks.lazydocker.Color
+---@field optionsTextColor snacks.lazydocker.Color
+---@field searchingActiveBorderColor snacks.lazydocker.Color
+---@field selectedLineBgColor snacks.lazydocker.Color
+---@field unstagedChangesColor snacks.lazydocker.Color
+
+---@class snacks.lazydocker.Config: snacks.terminal.Opts
+---@field args? string[]
+---@field theme? snacks.lazydocker.Theme
+local defaults = {
+  -- automatically configure lazydocker to use the current colorscheme
+  -- and integrate edit with the current neovim instance
+  configure = true,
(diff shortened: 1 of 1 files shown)


lua/snacks/lazydocker.lua
```diff
@@ -0,0 +1,212 @@
...
+    [241]                      = { fg = "Special" },
+    activeBorderColor          = { fg = "MatchParen", bold = true },
+    cherryPickedCommitBgColor  = { fg = "Identifier" },
+    cherryPickedCommitFgColor  = { fg = "Function" },
+    defaultFgColor             = { fg = "Normal" },
+    inactiveBorderColor        = { fg = "FloatBorder" },
+    optionsTextColor           = { fg = "Function" },
+    searchingActiveBorderColor = { fg = "MatchParen", bold = true },
+    selectedLineBgColor        = { bg = "Visual" }, -- set to `default` to have no background colour
+    unstagedChangesColor       = { fg = "DiagnosticError" },
```
folke, line 56: These colors don't exist in lazydocker

lua/snacks/lazydocker.lua
```diff
@@ -0,0 +1,212 @@
...
+  -- and integrate edit with the current neovim instance
+  configure = true,
+  -- extra configuration for lazydocker that will be merged with the default
+  -- snacks does NOT have a full yaml parser, so if you need `"test"` to appear with the quotes
+  -- you need to double quote it: `"\"test\""`
+  config = {
+    os = { editPreset = "nvim-remote" },
+    gui = {
+      -- set to an empty string "" to disable icons
+      nerdFontsVersion = "3",
```
folke, line 40: options don't exist

lua/snacks/lazydocker.lua
```diff
@@ -0,0 +1,212 @@
...
+        "Failed to get **lazydocker** config directory.",
+        "Will not apply **lazydocker** config.",
+        "",
+        "# Error:",
+        vim.trim(out),
+      }
+      Snacks.notify.error(msg, { title = "lazydocker" })
+    end
+  end
+end
```
folke, line 112: can't work with lazydocker

folke (commented):
I'll keep the review as is for now.

In general all of this simply doesn't work with lazydocker.

You can also just use `Snacks.terminal("lazydocker")`. with the lazygit style.

I Don't see why this shouold be a plugin since you can't actually configure anything or at least not like we do for lazygit.

umutondersu:
Thank you for your feedback. I did not know the snack.terminal api could be used this way. Should have been more careful reading the docs😅​

S1M0N38:
Screenshot 2025-10-21 at 09: 53: 50
Screenshot 2025-10-21 at 09: 52: 35


It's true that lazydocker does not have all the colors implemented in this PR, but some of them exist. Here is the default [lazydocker config](https://github.com/jesseduffield/lazydocker/blob/master/docs/Config.md#default)

So maybe it would be nice to set

```
activeBorderColor
inactiveBorderColor
selectedLineBgColor
optionsTextColor
```

with the same color used in lazygit for better color consistency across nvim/lazygit/lazydocker.

PR #1006, opened 2025-02-08
docs(rename): add oil.nvim integration

f-leeske:
## Description

Add an autocmd for oil.nvim integration of snacks.rename to the docs, similar to mini.files.


folke:
Great, ty!

PR #1019, opened 2025-02-08
feat(picker.scratch)

maskudo:

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -833,4 +833,36 @@ M.zoxide = {
   },
 }
 
+M.scratch = {
+  finder = "files_scratch",
+  format = "file",
+  win = {
+    preview = { wo = { number = false, relativenumber = false, signcolumn = "no" } },
+    input = {
+      keys = {
+        ["<CR>"] = { "scratch_open", mode = { "n", "i" } },
+        ["<c-x>"] = { "scratch_delete", mode = { "n", "i" } },
+        ["<c-n>"] = { "scratch_new", mode = { "n", "i" } },
+      },
+    },
+  },
+  actions = {
+    scratch_open = function(picker)
+      local current = picker:current().file
+      picker:close()
+      Snacks.scratch.open({ file = current })
+    end,
+    scratch_delete = function(picker)
+      local current = picker:current().file
+      os.remove(current)
+      picker:close()
+      Snacks.picker.scratch()
+    end,
+    scratch_new = function(picker)
+      picker:close()
+      Snacks.scratch.open()
+    end,
+  },
+}
+
 return M
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -843,6 +843,7 @@ M.scratch = {
         ["<CR>"] = { "scratch_open", mode = { "n", "i" } },
         ["<c-x>"] = { "scratch_delete", mode = { "n", "i" } },
         ["<c-n>"] = { "scratch_new", mode = { "n", "i" } },
+        ["<c-g>"] = { "scratch_grep", mode = { "n", "i" } },
       },
     },
   },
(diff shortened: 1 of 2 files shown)


dpetka2001:
I have the following for my custom scratch picker
```lua
function M.scratch_picker()
  ---@type snacks.picker.Config
  return {
    win = {
      input = {
        keys = {
          ["<c-d>"] = { "delete", mode = { "i", "n" } },
        },
      },
    },
    finder = function()
      local list = Snacks.scratch.list()
      local items = {} ---@type snacks.scratch.Item[]
      for _, item in ipairs(list) do
        items[#items + 1] = {
          file = item.file,
          item = item,
          title = item.name,
          text = Snacks.picker.util.text(item, { "name", "branch", "ft" }),
          branch = item.branch and ("branch:%s"):format(item.branch) or "",
        }
      end
      return items
    end,
    ---@param item snacks.scratch.Item
    format = function(item, picker)
      local file = item.item
      local ret = {} ---@type snacks.picker.Highlight[]
      local a = Snacks.picker.util.align
      local icon, icon_hl = file.icon, nil
      if not icon then
        icon, icon_hl = Snacks.util.icon(file.ft, "filetype")
      end
      ret[#ret + 1] = { a(icon, 3), icon_hl }
      ret[#ret + 1] = { a(file.name, 20, { truncate = true }) }
      ret[#ret + 1] = { " " }
      ret[#ret + 1] = { a(item.branch, 20, { truncate = true }), "Number" }
      ret[#ret + 1] = { " " }
      ---@diagnostic disable-next-line: missing-fields
      vim.list_extend(ret, Snacks.picker.format.filename({ text = "", dir = true, file = file.cwd }, picker))
      return ret
    end,
    actions = {
      ---@param item snacks.scratch.Item
      confirm = function(picker, item)
        picker:close()
        if not item then
          return
        end
        Snacks.scratch.open({ icon = item.item.icon, file = item.item.file, name = item.item.name, ft = item.item.ft })
      end,
      ---@param picker snacks.Picker
      ---@param item snacks.scratch.Item
      delete = function(picker, item)
        local deleted = {}
        local done = {}
        local name = vim.fn.fnamemodify(item.file, ":p:~:.")

        Snacks.notify("Deleted scratch file:\n- `" .. name .. "`", { title = "Scratch" })
        os.remove(item.file)
        deleted["idx"] = item.idx

        picker:find({
          on_done = function()
            if picker:count() == 0 then
              picker:close()
            else
              -- keep track of `idx` so the cursorline does not move to the top when you delete
              for _, it in ipairs(picker.list.items) do
                done[#done + 1] = it.idx
              end
              if not vim.tbl_contains(done, deleted.idx) then
                item.idx = item.idx - 1
              end
              picker.list:view(item.idx)
            end
          end,
        })
      end,
    },
  }
end
```
I believe it's unnecessary to use `proc` finder to use the `fd` command, since `Snacks.scratch.list()` already does this for you. You just have to transform the items in the finder. This was done with the help of Folke long time ago when `snacks.picker` first appeared.

Feel free to use whatever you want if you find it to your liking.

acidclouds:
This is awesome. Can this be merged?

bagaze:
not stale

bagaze:
not stale

carlesandres:
not stale.

carlesandres:
@folke it is not straightforward how to get rid of scratch buffers. 
This would help with that.

It also provides some other nice enhancements. Any chance you can take a look?

folke:
@dpetka2001 where can I find your latest code for the scratch picker?
Will look into this tomorrow, but would be good to see what you did as well. Thanks!

dpetka2001:
Latest code is what is used in this PR as well. Although I've only implemented `confirm` and `delete` actions for my purposes (don't have an abundance of scratch files around) and also did not change window options for not showing number and signcolumn like this PR does. 

So, 2 actions less and no changing window options is the only difference. Rest exactly the same.

```lua
-- confirm action
confirm = function(picker, item)
  picker:close()
  if not item then
    return
  end
  Snacks.scratch.open({ icon = item.item.icon, file = item.item.file, name = item.item.name, ft = item.item.ft })
end,
```
I have this instead of `scratch_open` action that this PR has. I just prefered `confirm` instead of custom action, because `<Cr>` is already mapped to it, so no need to define it in `win.input.keys`.

folke:
- Refactored the picker action code and scratch.
- Snacks.scratch.select() now just does Snacks.picker.scratch()
- removed the grep action. you can add this in your own config.
ty!

PR #1033, opened 2025-02-09
feat(picker.command_history): modify commad before executing

maskudo:
## Description

Currently, there is no way to modify selected command before executing. 
This PR adds keybind to accept the current selected item as completion so commands can be manipulated before executing. 

## Related Issue(s)

https://github.com/folke/snacks.nvim/discussions/929

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -144,6 +144,19 @@ M.command_history = {
   },
   confirm = "cmd",
   formatters = { text = { ft = "vim" } },
+  win = {
+    input = {
+      keys = {
+        ["<c-y>"] = { "accept_completion", mode = { "n", "i" } },
+      },
+    },
+  },
+  actions = {
+    accept_completion = function(picker)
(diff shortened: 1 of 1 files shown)


folke:
Your PR would not have worked. It just changes the filter, but pressing `<CR>` does not magically use the filter input.

I've just update snacks to always start the cmdline when confirming a cmd.

antonk52:
Can we separate default confirm actions for commands picker and command history picker or make this configurable? This adds an extra step or friction for a regular command picker

michaelPotter:
Here's my workaround to re-enable the "accept if nargs == 0" functionality...

``` lua
local pickerConfig = {

  sources = {
    commands = {
      confirm = "cmd0",
    },
  },

  actions = {
    cmd0 = function(picker, item)
      if item and item.command and item.command.nargs == "0" then
        picker:close()
        vim.schedule(function()
          vim.cmd(item.cmd)
        end)
      else
        require('snacks.picker.actions').cmd(picker, item)
      end
    end,
  },

}
```

PR #1089, opened 2025-02-11
fix(gitbrowse): add support for GitHub Enterprise Cloud repo url

brablc:
The extra % before - is needed in Lua. Without this rule, `<leader>gB` would just open empty browser.

## Description

Support repo url in format: `org-123456@github.com:repo-group/repo-name.git`


folke:
ty!

PR #1091, opened 2025-02-11
feat(git_log): add author filter

gogamid:
## Description

I was using LazyGit feature to filter git logs by the author, unfortunately it was missing in snacks picker. Here is my solution to it. Let me know if you have better suggestion :)


gogamid:
@folke  what do you think about this?

folke:
removed the example, fixed the order of the author arg and merged. ty!

PR #1093, opened 2025-02-12
fix(statuscolumn): only execute `za` when fold exists

felipeymn:
## Description

Currently, clicking the status column on a line without a fold results in an error (see screenshot below). This PR adds a check before executing `za` to prevent this issue.

## Screenshots
statuscolumn


folke:
ty

PR #1105, opened 2025-02-12
feat(treesitter): add `tree` boolean to toggle on/off tree symbols

dpetka2001:
## Description
Adds a `opts.tree` boolean option for treesitter picker to toggle on/off tree symbols

## Related Issue(s)
None, rather a discussion #1101


folke:
ty!

PR #1121, opened 2025-02-12
feat(toggle): allow user to add custom which-key description

### LINKED ISSUE #1114: feature(toggle): customizable "{En,Dis}able" text in which-key descriptions
### Did you check the docs?

Toggles with a mapped key are described as "{En,Dis}able <NAME>" in the which-key pop-up, and I'm slightly annoyed by the fact that the items don't line up because "Disable" is longer than "Enable". By padding "Enable" with an additional space, option names would perfectly line up!

Others might want to use a static "Toggle <NAME>" for consistency with other plugins. Or maybe some minimalists are content with just the toggle name, foregoing any prefix.

Adding a new config option, `which_key_prefix` with default value `{ enable = "Enable ", disable = "Disable " }`.

None

### Additional context

I could whip up a PR if there's interest.

dpetka2001:
## Description
Allows user to add custom which-key description in the `toggle` itself

## Related Issue(s)
Closes #1114

### DIFF
--- a/lua/snacks/toggle.lua
+++ b/lua/snacks/toggle.lua
@@ -13,6 +13,7 @@ M.meta = {
 ---@class snacks.toggle.Config
 ---@field icon? string|{ enabled: string, disabled: string }
 ---@field color? string|{ enabled: string, disabled: string }
+---@field wk_desc? string|{ enabled: string, disabled: string }
 ---@field map? fun(mode: string|string[], lhs: string, rhs: string|fun(), opts?: vim.keymap.set.Opts)
 ---@field which_key? boolean
 ---@field notify? boolean
@@ -29,6 +30,10 @@ local defaults = {
   color = {
     enabled = "green",
     disabled = "yellow",
+  },
+  wk_desc = {
+    enabled = "Disable ",
+    disabled = "Enable ",
   },
 }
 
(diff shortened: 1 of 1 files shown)


folke:
ty!

sjclayton:
@dpetka2001 

Is there any way to have this behaviour only apply to specific mappings, or is it always just all of them?

dpetka2001:
@sjclayton Or if you mean in the context of a Neovim distro (like LazyVim for example), which already has predefined keymaps and you just want to change the defaults for all keymaps already defined instead of having to do it for each specific mapping explicitly, you could also do the following
```lua
opts = {
	toggle = {
		config = function(opts, defaults)
			opts.wk_desc = {
				enabled = "test ",
				disabled = "test2 ",
			}
		end,
	},
}
```
Those will be the new defaults for every mapping you create if you don't specify `wk_desc` in a specific mapping explicitly.

sjclayton:
@dpetka2001 

I figured out how to do it for all of them, I just wasn't sure if I could do it on a per mapping basis by just passing that table directly to an individual call to `Snacks.toggle()` or not.

dpetka2001:
Then yes, you can do it how I mentioned in my initial comment with passing the table to `Snacks.toggle`. But remove that `config` function because from my testing it will overwrite your custom descriptions, because it gets evaluated after the `snacks.toggle` gets resolved.

PR #1123, opened 2025-02-13
feat(picker): add LSP symbol range to result item

### LINKED ISSUE #1057: feature(picker): focus current position for LSP symbols picker
### Did you check the docs?

When I launch `Snacks.picker.lsp_symbols()` it always jumps to the first document symbol

I would like the LSP symbols picker to focus the symbol at the cursor position, e.g. the class or method that I am currently in

https://github.com/SmiteshP/nvim-navbuddy does that by default

### Additional context

_No response_

disrupted:
## Description

this adds an attribute for the range of the LSP symbol to each picker item.

## Related Issue(s)

Fixes #1057 

having access to the range of the LSP symbol (i.e. the whole context of classes and methods) would be tremendously helpful for my solution in https://github.com/folke/snacks.nvim/issues/1057#issuecomment-2652052218. Currently the picker items only have a `loc` attribute which holds the range of the symbol identifier, i.e. the class/method name.

---

perhaps in a followup we could use the range to highlight the symbol context, similar to how trouble.nvim does it

### DIFF
--- a/lua/snacks/picker/source/lsp/init.lua
+++ b/lua/snacks/picker/source/lsp/init.lua
@@ -260,6 +260,7 @@ function M.results_to_items(client, results, opts)
       detail = result.detail,
       name = result.name,
       text = "",
+      range = result.range,
     }
     local uri = result.location and result.location.uri or result.uri or opts.default_uri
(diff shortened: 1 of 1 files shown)


folke:
ty!

PR #1127, opened 2025-02-13
fix(gitbrowse): previous logic always overwrote 'commit'

dpetka2001:
## Description
The previous logic didn't check for `fields.commit` when checking whether to switch to `branch` or `repo`, thus `opts.what = "commit"` was always being overwritten.

## Related Issue(s)
None


folke:
thanks again! :)

PR #1131, opened 2025-02-13
ci(labeler): add picker to labeler

marcinjahn:
## Description

There are quite a few PRs relating to picker, and they are not labeled.


folke:
ty

PR #1145, opened 2025-02-13
fix(picker.explorer): ensure diagnostics can be disabled

disrupted:
## Description

`opts.diagnostics=false` was being ignored by `Snacks.picker.explorer`

## Related Issue(s)

I didn't create an issue for it


folke:
Great, ty!

PR #1164, opened 2025-02-14
fix(input): add missing hl group for input title

SheffeyG:
Add the missing highlight group for input title.
Before:
image
After:
image


lua/snacks/input.lua
```diff
@@ -151,7 +151,7 @@ function M.input(opts, on_confirm)
   add(opts.prompt, "SnacksInputTitle", opts.prompt_pos)
 
   if next(title) then
-    table.insert(title, { " " })
+    table.insert(title, { " ", "SnacksInputTitle" })
```
folke, line 154: Should be `FloatTitle`

folke:
nope, all correct. ty!

PR #1171, opened 2025-02-14
feat(picker.lsp): add new option `lsp_result_transform`

fang2hou:
## Description

To extend the functionality of the picker for processing LSP symbols, I added the `lsp_result_transform` option, which allows further modification of picker items with results returned by LSP.

I created a example with this option, see: https://github.com/fang2hou/go-impl.nvim/pull/1.

In the above case, the Go Language Server `gopls` returns a `containerName` field that indicates the package where the symbol is located. Currently, if I want to retrieve the package name of a symbol, I need to clone the built-in snacks finder function and make a slight modification to include the `containerName` field during post-processing.

Since many LSPs return custom fields based on their own language specifications, variations like this may quite common. Adding an option makes it easier for plugin authors to integrate with the snacks.picker without worrying about differences between the built-in `lsp_symbols` finder and a cloned version.

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -539,6 +539,7 @@ M.lsp_references = {
 ---@field tree? boolean show symbol tree
 ---@field filter table<string, string[]|boolean>? symbol kind filter
 ---@field workspace? boolean show workspace symbols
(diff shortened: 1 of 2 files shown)


folke:
`item.item` is now the original result.
You can use the exiting `opts.transform` functionlity for the rest. no need for a new option

PR #1187, opened 2025-02-15
Start mode option to Snacks Input

mbenford:
This PR adds a new option to Snacks Input, `start_mode`, to set the initial mode when the input window opens, inspired by [Dressing.nvim](https://github.com/stevearc/dressing.nvim). Changing the initial mode can be useful when a default value is used (e.g. LSP Rename) and that value is expected to be edited.


folke:
not interested in adding this

PR #1199, opened 2025-02-15
feat(picker.explorer): support custom icon for open directory

xoumi:
## Description

The icon used for open directories in explorer picker  is hardcoded to a folder icon from the fontawesome library (nerdfont name: `nf-fa-*`) but the ones generally used by `mini.icons` are from material design (nerdfont name: `nf-md-*`). 

This PR adds support for custom icon for open directories and sets the default open directory icon to that provided my material design.

## Screenshots
Before:
Screenshot 2025-02-15 at 7 58 46 PM
After:
Screenshot 2025-02-15 at 8 00 32 PM

### DIFF
--- a/doc/snacks-picker.txt
+++ b/doc/snacks-picker.txt
@@ -504,6 +504,7 @@ below.
           vertical = "│ ",
           middle   = "├╴",
           last     = "└╴",
+          open_dir = "󰝰 ",
         },
         undo = {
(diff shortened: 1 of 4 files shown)


folke:
changed the default to md and fixed it in a couple of other places as well.
icons can be configured with `opts.icons.files.dir/dir_open/file`

PR #1200, opened 2025-02-15
fix(filter): Fix filtering condition check

tsakirist:
Check the boolean field of the filter path only when there is a direct
match. Otherwise a filter's path boolean value that is true dominates
the results.

Example:

```lua
opts = {
  sources = {
    recent =  {
      filter = {
        paths = {
          [vim.fn.stdpath("data")] = true,
        }
      }
    }
  }
}
```

This would cause the recent picker to only include paths from the
stdpath("data"). Since the previous condition could never allow any
other path to return true.

But I merely want to enable that path to be included in the results.

### DIFF
--- a/lua/snacks/picker/core/filter.lua
+++ b/lua/snacks/picker/core/filter.lua
@@ -101,8 +101,8 @@ function M:match(item)
   end
   if self.opts.paths then
     for _, p in ipairs(self.paths) do
-      if (path:sub(1, #p.path) == p.path) ~= p.want then
-        return false
+      if path:find("^" .. p.path) then
+        return p.want
       end
     end
   end


tsakirist:
Ok, so how would I go about disabling that default filter in recents to also include results from that path? Unless I'm missing something it's impossible now

tsakirist:
Hi @folke,

Is it currently possible to disable the default recent files picker filter paths?
e.g. allowing results from `[vim.fn.stdpath("data")]` ?
I would like to avoid creating a new custom picker for this.

Thanks.

tkapous:
Stumbled on the same problem, [here](https://gitlab.com/Biggybi/neovim-config/-/commit/4b41119)'s how I did it.

PR #1216, opened 2025-02-16
fix(picker.git): preserve chronological order when matching

pkazmier:
## Description

When using the git log sources, what are your thoughts on preserving the sort order when matching? I find most often when I use this picker I want to see results in reverse chronological order. For example, I like to use the picker to review the commit log to see new features added to a repo.

## Screenshots

When matching for `feat` in the snacks repo with current behavior, the output is less than useful:
image

But with the proposed change, it's much more relevant:
image

## Alternatives

I can override in my config:

```lua
      picker = {
        sources = {
          git_log = { sort = { fields = { "score:desc", "idx" } } },
          git_log_file = { sort = { fields = { "score:desc", "idx" } } },
          git_log_line = { sort = { fields = { "score:desc", "idx" } } },
        },
      },
```

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -270,6 +270,7 @@ M.git_log = {
   format = "git_log",
   preview = "git_show",
   confirm = "git_checkout",
+  sort = { fields = { "score:desc", "idx" } },
 }
 
 ---@type snacks.picker.git.log.Config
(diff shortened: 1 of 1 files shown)


folke:
Makes sense. Thank you!

lalvarezt:
alternative option here, in case anyone is interested
https://github.com/folke/snacks.nvim/discussions/1243#discussioncomment-12228711

PR #1250, opened 2025-02-17
perf(dashboard): speed up filtering for recent_files

joshuali925:
## Description

I have 5000 oldfiles, every time i go to a new directory and open nvim, the dashboard recent_files (with `cwd=true`) computation takes 4-5 seconds to go through all oldfiles before deciding nothing should show up

```lua
sections = {
  { icon = " ", title = "Recent files (current directory)", section = "recent_files", cwd = true },
}
```

It appears to be the `vim.v.oldfiles[i]` indexing. I didn't look much into it but it opens much faster looping with ipairs.

Repro:
1. add the above section config in dashboard
2. increase oldfiles limit and populate up to the limit
3. `mkdir brand-new-directory; cd brand-new-directory; nvim`
4. the section is empty as expected, but it takes a few seconds before dashboard shows up

### DIFF
--- a/lua/snacks/dashboard.lua
+++ b/lua/snacks/dashboard.lua
@@ -796,9 +796,10 @@ function M.oldfiles(opts)
   end
   local done = {} ---@type table<string, boolean>
   local i = 1
+  local oldfiles = vim.v.oldfiles
   return function()
-    while vim.v.oldfiles[i] do
-      local file = vim.fs.normalize(vim.v.oldfiles[i], { _fast = true, expand_env = false })
+    while oldfiles[i] do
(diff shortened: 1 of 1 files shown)


folke:
The only problem is that your code is **completely** wrong now.
That returns a generator, but right now it will loop over all data every time.
Really weird that this is faster.
This will no longer do what it needs to do, unless I'm missing something?

folke:
If indexing of oldfiles is an issue, then put it in a separate list before the function() and use the new list inside the function with the original (correct) code.

joshuali925:
thanks for taking a look. i see what you mean, i missed that `i` is outside of the scope. Although it didn't impact correctness and was significantly faster.

i updated the PR

folke:
ty!

PR #1272, opened 2025-02-18
feat(picker.format): abbrevation options for file formatter

ADIX7:
## Description

The file formatter currently strips all the folders from the path if the whole path is too long. When there are multiple files with the same name the user can only choose based on the preview (in case it is present).

This PR adds and options to abbreviate the folder names similarly how some file explorer does that (I'm sure I saw it, but I can't find any example...). The default is false for this, so the current behaviour is kept by default.

The abbreviation will get same length substring from the beginning of every folder. It will take as many characters as it can by not exceeding the truncate limit.

Note that there is no `...` when a folder is abbreviated so there is no obvious sign if a folder name has been abbreviated. The logic behind this is that it would take even more space, although an option can be added for this.

## Screenshots

kép


ADIX7:
The current implementation does not abbreviate the first folder. I'm not sure if this behaviour should be kept or it should be abbreviated as well.

folke:
not interested in adding this. This PR also removes a bunch of files which is obviously not mergeable

Issue #1291, opened 2025-02-18
bug(picker): Can't correctly select text like \<foo\> from search_history picker


metziger23:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4 Build type: Release LuaJIT 2.1.1736781742

### Operating system/version

Macos 15.2

### Describe the bug

If I select a text  like `\<foo\>` from search_history picker, the `\` symbol is pasted inside my search cmdline instead of `\<foo\>`
So this basically means that I'm unable to restore the searches that I got by using `*` on  the current text. I also tested this on nvim nightly but the result is the same.

https://github.com/user-attachments/assets/c8875fb2-dd76-4e55-ac8c-9bbfabb97401

### Steps To Reproduce

1. Select any text using `*` (trigger search of the text under the cursor)
2. Open require("snacks").picker.search_history()
3. Select the text that was put in search history by pressing `*`

### Expected Behavior

text like \<foo\> should appear in search cmdline instead of `\`

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

---@diagnostic disable-next-line: missing-fields
require("lazy.minit").repro({
	spec = {
		{
			"folke/snacks.nvim",
			opts = {
				picker = {},
			},
			keys = {
				{
					"<space>",
					function()
						require("snacks").picker.search_history()
					end,
					desc = "Search History",
				},
			},
		},
		-- add any other plugins here
	},
})
```

metziger23:
It works correctly now. Thank you for the fix!

PR #1292, opened 2025-02-18
fix(picker.actions): `search` action correctly treat `<`

dpetka2001:
## Description
`nvim_input` treats `<` character as special, so pass `<lt>` for literal `<` if found in `item.text`.

## Related Issue(s)
#1291

### DIFF
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -495,7 +495,7 @@ end
 function M.search(picker, item)
   picker:close()
   if item then
-    vim.api.nvim_input("/" .. item.text)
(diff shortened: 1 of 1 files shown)


folke:
I just fixed something else there, so you'll have to rebase. Ty!

dpetka2001:
I don't think it's needed any more. Because `item.text` is being set with `setcmdline` instead of `nvim_input` and it seems to be working correct now. Could you also test on your side and if the linked issue is resolved, then just close this?

dpetka2001:
Yes, I just saw the issue and the change you made was for the same issue. So, I'm closing this.

folke:
ok, even better!

PR #1296, opened 2025-02-19
feat(words): add `filter` function for user to disable specific filetypes

### LINKED ISSUE #1294: feature: words should provide per-filetype disable option
### Did you check the docs?

We should be able to selectively disable `Snacks.words` in certain buffers or filetypes (e.g., if the corresponding LSP is slow).

This could be achieved e.g. with a global `Snacks.words.exclude_filetypes` option.

Alternatively, providing a buffer-local `vim.b.snacks_words_enable` option (similar to `vim.b.snacks_animate`) might work, e.g. by setting it on a `FileType` autocmd (I think).

We can manually toggle `Snacks.words` globally when we open or navigate into or out of a problematic buffer. However, this is very tedious and annoying.

### Additional context

_No response_

dpetka2001:
## Description
This allows user to specify explicitly filetypes to disable `snacks.words`.

## Related Issue(s)
Closes #1294

### DIFF
--- a/lua/snacks/words.lua
+++ b/lua/snacks/words.lua
@@ -18,6 +18,7 @@ local defaults = {
   foldopen = true, -- open folds after jumping
   jumplist = true, -- set jump point before jumping
   modes = { "n", "i", "c" }, -- modes to show references
+  exclude_ft = {}, -- filetypes to exclude
 }
 
 M.enabled = false
@@ -95,6 +96,13 @@ function M.is_enabled(opts)
     mode = mode:sub(1, 2) == "no" and "o" or mode
     mode = mode:sub(1, 1):match("[ncitsvo]") or "n"
     if not vim.tbl_contains(config.modes, mode) then
+      return false
+    end
+  end
+
+  if not vim.tbl_isempty(config.exclude_ft) then
+    local ft = vim.bo[vim.api.nvim_get_current_buf()].filetype
+    if vim.tbl_contains(config.exclude_ft, ft) then
       return false
     end
   end
--- a/lua/snacks/words.lua
+++ b/lua/snacks/words.lua
@@ -18,7 +18,9 @@ local defaults = {
   foldopen = true, -- open folds after jumping
   jumplist = true, -- set jump point before jumping
   modes = { "n", "i", "c" }, -- modes to show references
-  exclude_ft = {}, -- filetypes to exclude
+  filter = function(buf) -- what buffers to enable `snacks.words`
+    return vim.g.snacks_words ~= false and vim.b[buf].snacks_words ~= false
+  end,
 }
 
 M.enabled = false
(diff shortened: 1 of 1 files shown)


lua/snacks/words.lua
```diff
@@ -99,6 +100,13 @@ function M.is_enabled(opts)
...
   end
 
+  if not vim.tbl_isempty(config.exclude_ft) then
+    local ft = vim.bo[vim.api.nvim_get_current_buf()].filetype
+    if vim.tbl_contains(config.exclude_ft, ft) then
+      return false
+    end
+  end
+
   local buf = opts.buf or vim.api.nvim_get_current_buf()
```
tristan957, line 110: Should you use the buf variable declared here?

folke:
Can your do it like we do for scroll / indent and others instead?

https://github.com/folke/snacks.nvim/blob/main/lua/snacks/scroll.lua#L37-L41

No need to check for buftype in this case though

lua/snacks/words.lua
dpetka2001: I didn't think `opts.buf` would ever be used, because we only cared about the current buffer. Either way I changed the implementation according to Folke's suggestion.

dpetka2001:
Done.

dpetka2001:
Just curious, but if you don't mind me asking, what's the benefit of using the latest implementation over the previous one? Just a personal preference and for implementation uniformity like with the others or is there some other benefit I fail to think of?

lua/snacks/words.lua
```diff
@@ -18,7 +18,9 @@ local defaults = {
   foldopen = true, -- open folds after jumping
   jumplist = true, -- set jump point before jumping
   modes = { "n", "i", "c" }, -- modes to show references
-  exclude_ft = {}, -- filetypes to exclude
+  filter = function(buf) -- what buffers to enable `snacks.words`
+    return vim.g.words_disable ~= false and vim.b[buf].words_disable ~= false
```
folke, line 22: can you change this to `vim.g.snacks_words ~= false ...`?
We use the same naming scheme in other places

dpetka2001: Yes, done. Sorry I didn't think about naming scheme.

folke:
> Just curious, but if you don't mind me asking, what's the benefit of using the latest implementation over the previous one? Just a personal preference and for implementation uniformity like with the others or is there some other benefit I fail to think of?
> 
> PS: one drawback might be that with the previous implementation the autocmd had to be in `LspAttach` (because that's when `snacks.words` is set up and `config.exclude_ft` wouldn't be available), whereas with the buffer-local variable you can just do it in `FileType` from my testing (because even when the plugin is set up it just checks the variable).

just for uniformity. best to use the same way to achieve the same thing accross plugins.

folke:
Looks good, ty!

Issue #1258, opened 2025-02-18
bug: `Snacks.image.hover()` behaves weird


yuukibarns:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

10.4

### Operating system/version

Arch Linux x86_64

### Describe the bug

https://github.com/user-attachments/assets/13816d30-413d-4330-ae88-f456d60b57d9

1. image preview is not corresponding to inline math under cursor, instead, it is corresponding to the first inline math of current line.
2. image preview keeps showing up after cursor movement.

I find the implementation of `Snacks.image.hover()` parsing the current line under the cursor and the line after,
```lua
-- line 
--- Get the image at the cursor (if any)
---@return string? image_src, snacks.image.Pos? image_pos
function M.at_cursor()
  local cursor = vim.api.nvim_win_get_cursor(0)
  local img = M.find(vim.api.nvim_get_current_buf(), cursor[1], cursor[1] + 1)[1]
  if img then
    return img.src, img.pos
  end
end
```
I do not know if it is feasible to locate the image more precisely?

And if I understand it correctly, the implementation has put the cursor movement into consideration, but it does not work properly?
```lua
  vim.api.nvim_create_autocmd({ "BufWritePost", "CursorMoved", "ModeChanged", "BufLeave" }, {
    group = vim.api.nvim_create_augroup("snacks.image.hover", { clear = true }),
    callback = function()
      if not hover then
        return true
      end
      M.hover()
      if not hover then
        return true
      end
    end,
  })
```

### Steps To Reproduce

```lua
return {
	{
		"folke/snacks.nvim",
		priority = 1000,
		keys = {
			{
				"<leader>mp",
				ft = { "tex", "markdown" },
				function()
					Snacks.image.hover()
				end,
				desc = "Preview math formula under cursor"
			}

		},
		---@type snacks.Config
		opts = {
			image = {
				enabled = true,
				doc = {
					-- enable image viewer for documents
					-- a treesitter parser must be available for the enabled languages.
					-- supported language injections: markdown, html
					enabled = true,
					-- render the image inline in the buffer
					-- if your env doesn't support unicode placeholders, this will be disabled
					-- takes precedence over `opts.float` on supported terminals
					inline = false,
[...]

PR #1298, opened 2025-02-19
feat(image): add `Snacks.image.hover_inline()` function

yuukibarns:
## Description

https://github.com/folke/snacks.nvim/issues/1258#issuecomment-2665210723

When `inline = true` there are too many images, `Snacks.image.hover()` does not persist for afterwards reference.

## Screenshots

[Screencast_20250219_113101.webm](https://github.com/user-attachments/assets/44fef140-930d-4e03-b227-231e6465b74e)

### DIFF
--- a/lua/snacks/image/doc.lua
+++ b/lua/snacks/image/doc.lua
@@ -198,7 +198,7 @@ function M.hover_close()
 end
 
 --- Get the image at the cursor (if any)
----@return string? image_src, snacks.image.Pos? image_pos
+---@return string? image_src, snacks.image.Pos? image_pos, string? image_id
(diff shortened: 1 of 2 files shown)


yuukibarns:
`inline = true` there are too many images, `Snacks.image.hover()` does not persist for afterwards reference.

yuukibarns:
I am not familiar with lua, and this PR may be too naive to merge. But I will appreciate it if you can take it into consideration and provide some similar api functions.

yuukibarns:
But after editing the buffer, treesitter nodes' ids will change which leads to previous opened images closing.

PR #1313, opened 2025-02-19
fix(picker.resume): propagate opts

Kirens:
## Description

This patch ensures `Snacks.picker.resume(opts)` propagate the provided opts. Currently, they are silently discarded even though the documentation and types indicate otherwise

 https://github.com/folke/snacks.nvim/blob/cfddd88376f7cf8626c5686ab5690b80010d667f/lua/snacks/picker/init.lua#L10
https://github.com/folke/snacks.nvim/blob/cfddd88376f7cf8626c5686ab5690b80010d667f/lua/snacks/picker/types.lua#L54

## Options
This seems to work. However, I don't know the implications of replacing `last.opts` with a copy in the `M.resume()`. Another suggestion is to update the types to reflect the current behavior.

### DIFF
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -95,7 +95,7 @@ function M.new(opts)
   self.init_opts = opts
   self.opts = Snacks.picker.config.get(opts)
   if self.opts.source == "resume" then
-    return M.resume()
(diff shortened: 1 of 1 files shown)


folke:
sorry but this makes little sense.
resume is to resume an existing picker, not mix it with new options.

PR #1320, opened 2025-02-19
fix(image.doc): crop inline typst equations properly

stefanboca:
## Description

Tall inline typst equations were previously cropped too much vertically. The change to the template now resize text frames to the bounding box of their glyphs. See https://typst.app/docs/reference/text/text/#parameters-top-edge for more details.

## Screenshots
Before:
image
After:
image


folke:
ty!

PR #1331, opened 2025-02-20
Allow debug evaluation of block selections

jason0x43:
## Description

This PR adds basic support for visual block selections to `Snacks.debug.run()`. This is useful when code doesn't start at the beginning of a line, such as in comment blocks.

## Screenshots

Screenshot 2025-02-19 at 23 02 35


folke:
ty

Issue #1192, opened 2025-02-15
refactor: replace `require('snacks.dashboard').icon` with `Snacks.util.icon`


drowning-cat:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

This change allows setting icons for both the dashboard and pickers by overriding a single function, `Snacks.util.icon`.

### Describe the solution you'd like

```lua
---@class snacks.dashboard.Config
local defaults = {
  formats = {
    icon = function(item)
      if item.file and item.icon == "file" or item.icon == "directory" then
        -- return M.icon(item.file, item.icon)
        local icon, hl = Snacks.util.icon(item.file, item.icon)
        return { icon, width = 2, hl = hl }
      end
      return { item.icon, width = 2, hl = "icon" }
    end,
  }
}
```

### Describe alternatives you've considered

Remove `require('snacks.dashboard').icon` entirely.

### Additional context

Partially related:
 - https://github.com/folke/snacks.nvim/issues/558#issuecomment-2597016398
 - https://github.com/folke/snacks.nvim/discussions/1170

drowning-cat:
It might also be a good idea to separate icon pack calls into functions. This will allow for an easier setup (https://github.com/folke/snacks.nvim/issues/558)

Example:
```lua
Snacks.util.icon = Snacks.util.get_nvim_web_devicons
Snacks.util.icon = Snacks.util.get_mini_icons
```

Issue #558, opened 2025-01-16
feature: Allow to choose nvim-dev-icons over mini.icons even with both installed


dansdantas:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

Right now, because of how icons are fetched from Snacks.util.icon, if have both plugins are installed, mini-icons takes priority.

### Describe the solution you'd like

A configuration option to select which icon to use.

### Describe alternatives you've considered

If I disable mini.nvim, nvim-web-devicons shows correctly. Mini.icons cannot be disable by itself as other mini plugins.

folke:
Just override `Snacks.util.icon`. No need to proivde a config option for this.

PR #1351, opened 2025-02-21
refactor(util): extract `icon_from_nvim_web_devicons()` from `icon()`

drowning-cat:
## Description

- [x] Allow changing the default icon to `nvim-web-devicons` using:
```lua
Snacks.util.icon = Snacks.util.icon_from_nvim_web_devicons
```
- [x] Ensure the same fallback for `snacks.dashboard` and `snacks.util`. The fallback differs when using `nvim-web-devicons` for files without an extension (see ##Additional Comments).

## Related Issue(s)

https://github.com/folke/snacks.nvim/issues/1192#issuecomment-2660941500
https://github.com/folke/snacks.nvim/issues/558

## Screenshots

None

## Additional comments

> [!NOTE]
> `Snacks.util.icon_from_nvim_web_devicons` does return a fallback icon.

This means `snacks.dashboard.icon()` will no longer default to " " `nf-fa-files_o` (see [snacks/dashboard.lua#L746](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/dashboard.lua#L746)).
Instead, it will use `nf-md-file`, as `snacks.icon` now always returns a fallback when using `nvim-web-devicons`.

The fallback in `snacks.dashboard.icon()` can be safely removed (not done in this PR).

### DIFF
--- a/lua/snacks/util/init.lua
+++ b/lua/snacks/util/init.lua
@@ -102,10 +102,39 @@ function M.winhl(...)
   return Snacks.config.merge(unpack(ret))
 end
 
+---@alias snacks.util.IconParamOpts { fallback?: {dir?:string, file?:string} }
+---@alias snacks.util.IconFun fun(name: string, cat?: 'file'|string, opts?: snacks.util.IconParamOpts): string, string?
+
+local icon_fallback_file = "󰈔 "
+
+--- Get an icon from `nvim-web-devicons`.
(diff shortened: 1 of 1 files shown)


folke:
What an awful name for a function.

drowning-cat:
Yep, at the very least, `Snacks.util.icon_devicons` would have been better.

PR #1371, opened 2025-02-22
Support git browse from git log commit directly.

rjshrjndrn:
## Description

Right now there is not way to directly open a commit in browser from git log apart from checking out locally.
from any commit press `gb` in normal mode to open the commit in web.

## Screenshots

image

image

I'm using LazyVim and I've added the keybinding config in my plugin config. I think it'd be good if we can add as the default keybinding

```lua
return {
  "folke/snacks.nvim",
  opts = function(_, opts)
    local gitbrowse = require("snacks.gitbrowse")
    return vim.tbl_deep_extend("force", opts or {}, {
      picker = {
        sources = {
          -- Keybinding `gb` only for the git_log picker
          git_log = {
            win = {
              input = {
                keys = {
                  ["gb"] = {
                    "git_browse",
                    mode = { "n" },
                  },
                },
              },
            },
          },
        },
        actions = {
          git_browse = function(picker, item)
            -- if the picker have to be closed after the action, enable below.
            -- picker:close()
            gitbrowse.open({ what = "commit", commit = item.commit })
          end,
        },
      },
    })
  end,
}
```

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -15,6 +15,7 @@ local uv = vim.uv or vim.loop
 ---@class snacks.gitbrowse.Config
 ---@field url_patterns? table<string, table<string, string|fun(fields:snacks.gitbrowse.Fields):string>>
 local defaults = {
+  commit = nil, ---@type string
   notify = true, -- show notification on open
   -- Handler to open the url in a browser
   ---@param url string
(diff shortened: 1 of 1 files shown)


dpetka2001:
If you press `<M-w>` to cycle through to the list window, then if you put your cursor over a commit sha and press `<leader>gB` it will do what you want.

rjshrjndrn:
HI @folke Can you kindly have a look at the PR and suggest any changes needed.

Thank you.

rjshrjndrn:
@folke Gentle reminder

PR #393, opened 2024-12-26
fix/gitbrowse-commit

mosheavni:
## Description

Open a commit if cword is not a commit and `what` is `"commit"`

## Related Issue(s)

https://github.com/folke/snacks.nvim/discussions/373

PR #1381, opened 2025-02-22
feat(gitbrowse): get commit hash from current line on open

whoop-t:
## Description

Allow gitbrowse to get the current line commit hash when using `Snacks.gitbrowse.open()`

## Related Issue(s)

Currently you must hover the word of the actually commit hash. This requires you opening git blame(or something similar) and hovering the word of the commit hash you want to open.

I feel like this is kind of clunky and would it would be nicer to just be able to open commit url straight from file/line.

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -170,7 +170,7 @@ function M._open(opts)
     local word = vim.fn.expand("<cword>")
     fields.commit = is_valid_commit_hash(word, cwd) and word or nil
 
-    if not fields.commit then
+    if not fields.commit and not vim.fn.mode():find("[vV]") then
       local line_number = vim.fn.line('.')
       local blame_output = vim.fn.system(string.format("git blame -L %d,%d %s", line_number, line_number, file))
       local commit_hash = blame_output:match("^(%w+)")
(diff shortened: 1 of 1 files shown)


whoop-t:
Similar to: https://github.com/folke/snacks.nvim/pull/393

dpetka2001:
https://github.com/folke/snacks.nvim/blob/70e7e081ee558eb3756aba02491f1bc84fb72ab0/lua/snacks/gitbrowse.lua#L192-L200

These will never execute if you always set `fields.commit`. It's supposed to automatically switch the `opts.what` based on the content under the cursor without the user having to explicitly specify it.

#1371 will expose `opts.commit` and then users can implement this logic in their personal configuration if they want to.

whoop-t:
@dpetka2001 I guess then what's the point of setting `what = "commit"` if it just does stuff automatically.

Why not have another option like `auto` or something?

Without looking at the code it's kinda unclear what it will do when setting it to `commit`

dpetka2001:
`what = "commit"` is just the starting point and based on the context under the cursor it changes automatically if needed or not. You're supposed to put your cursor somewhere and based on that the appropriate action is taken. You're supposed to put the cursor on top of a commit sha if you want to trigger that action.

Anyhow, maintainer has the final say. I'm just a simple user.

whoop-t:
@dpetka2001 Its all good, this isnt super important but thought it was worth a discussion.

No where does it say you need to hover a commit sha for `what = "commit"`(the default setting) to work(unless you look at the code) so I think that will cause confusion. Im just coming from another plugin that if you say "i want to open the commit", thats what it does.

Maybe just a note somewhere as this confused me till I looked at the code.

Thanks for the back and forth

Issue #1369, opened 2025-02-21
bug: Latex preview


saxon1964:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.4

### Operating system/version

MacOS 15.3.1

### Describe the bug

I use default latex preview settings but instead of equations preview I often get just empty blocks where the PNG image should probably be. I have checked ~/.cache/nvim/snacks/image folder and all PNG files are there, perfect size and colors. But in Kitty, in .tex document, I don't see anything. Sometimes the images (not all of them) will show up if I change somethind and :w (write the file without closing). However, When I exit the document and re-enter it again, I don't get anything, even empty blocks of space are missing

### Steps To Reproduce

Use the following document and save as test.tex file:

`\documentclass{article}

\usepackage{amsmath, amssymb, amsfonts, amscd}

\begin{document}

\section{Equations}

$$z^n=x^n+y^n$$

Euler's most beautiful equation ever: $e^{i\pi}+1=0$
\[ \Gamma(x) = \int_0^\infty t^{x-1} e^{-t} dt \]

And now we are testing more equations:

\begin{equation}
  e^x=\sum_{n=0}^\infty \frac{x^n}{n!}
\end{equation}
\[
  \int_0^\infty e^{-x^2} dx = \sqrt{\pi}
\]

\[
  ax^2 + bx + c = 0 \implies x_{1,2} = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}  
\] 

\[
  \sum_{n=1}^\infty \frac{1}{n^2} = \frac{\pi^2}{6}
\]

\end{document}`

Open in neovim with default snacks.image settings: image={}. Check if you can see equation previews.

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/7003a137-aa18-4a4a-81d5-2392bd9ede33" />

### Expected Behavior

Equation previews shoudl be automatically rendered. PNG files are in the cache folder but kitty does not display them.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
[...]

PR #1396, opened 2025-02-23
feat(image): fixing problem with latex rendering in kitty

saxon1964:
Hi Folke,

A few days ago I reported a problem with latex rendering in kitty (bug: Latex preview #1369). you closed it quickly because you could not reproduce it, but I managed to do so every single time. Checkhealth report for Snacks on my Mac with Neovim is all green. So i decided to dig deeper into the code and in image/terminal.lua I have found the root cause of my problems.

In that file you have a hack:

 -- HACK: ghostty doesn't like it when sending images too fast,
 -- after Neovim startup, so we delay the first image

After that you queue data for 100ms, and when that period expires all buffered data is sent to stdout, buffered is deleted and after that everything goes directly to stdout. Fair enough, but that async aspect is something that kitty does not like, **like you have some kind of racing condition elsewhere.**

I tried to leave your hack as it is and to change the write function so that buffering is used only for ghostty in the following way:

```
 function M.write(data)
  if queue and M.env().name == "ghostty" then
    table.insert(queue, data)
  else
    io.stdout:write(data)
end
```

So buffering is enabled only for ghostty and for other terms writing is performed without buffering.

The situation improved immediately but in long latex documents (20+) equations I stll had an empty block here and there.

The solution that I'm submitting as a pull request is the only one that works for kitty 100% (I have spent the whole day testing) but still keeps that delay of 100ms needed by ghostty.

Basically, I ditched your buffering hack completely and modified the write function as follows.

```
local firstWrite = true

function M.write(data)
  if firstWrite then
    vim.wait(100)
    firstWrite = false
  end
  io.stdout:write(data)
end
```

Please, don't kill me for using vim.wait(100) - yes UI will be blocked for 0.1 seconds but I will personally buy a new Mac to everybody who notices that his computer was blocked for such a short time. So now everything is synchronous in terminal.lua and all problems have disappeared. Yes, it's a hack, a bit dirtier than yours, but it works.

### DIFF
--- a/lua/snacks/image/terminal.lua
+++ b/lua/snacks/image/terminal.lua
@@ -53,13 +53,13 @@ vim.api.nvim_create_autocmd("VimResized", {
 
 -- HACK: ghostty doesn't like it when sending images too fast,
 -- after Neovim startup, so we delay the first image
-local queue = {} ---@type string[]?
-vim.defer_fn(function()
-  if queue and #queue > 0 then
-    io.stdout:write(table.concat(queue, ""))
-  end
-  queue = nil
-end, 100)
+-- local queue = {} ---@type string[]?
+-- vim.defer_fn(function()
+--   if queue and #queue > 0 then
+--     io.stdout:write(table.concat(queue, ""))
+--   end
+--   queue = nil
+-- end, 100)
 
 function M.size()
   if size then
@@ -190,12 +190,14 @@ function M.set_cursor(pos)
   M.write("\27[" .. pos[1] .. ";" .. (pos[2] + 1) .. "H")
 end
 
+local firstWrite = true
+
 function M.write(data)
-  if queue then
-    table.insert(queue, data)
-  else
-    io.stdout:write(data)
+  if firstWrite then
+    vim.wait(100)
+    firstWrite = false
   end
+  io.stdout:write(data)
 end
 
 return M


folke:
Are you sure this is still an issue? I fixed it this morning: https://github.com/folke/snacks.nvim/commit/2b34c4dc05aa4cbccc6171fa530e95c218e9bc9c

Problem is simply that the queue never got reset in some cases.

That `vim.wait` is for sure a no-go.

saxon1964:
Ok, you know much better what's going on in the code. But bug #1369 was, after all, a genuine one. And your fix makes sense, I'm going to test it right away. I am just not sure why you are forcing delay for all terminals if only one is affected.

saxon1964:
Heh... It looks like 50% of the problem is solved. 

1) Delete all the files from cache directory: ~/.cache/nvim/snacks/image
2) nvim test.tex

Voila. It works. All equations rendered properly, cache dir populated.

Now exit nvim and run step (2) again: NOTHING gets rendered, image at the top is missing, all equations are represented by empty blocks. And the cursor is STUCK in the top left corner.

Delete the cache and everything works again.

I managed to fix the problem by chainging the write() function in the following way:

```
function M.write(data)
  if queue and M.env().name == "ghostty" then
    table.insert(queue, data)
  else
    io.stdout:write(data)
  end
end
```

folke:
Just to be clear, you only see this issue with kitty currently right? In Ghostty it works as expected?
I'll try to see if I can fix the actual issue.
The real problem is that Neovim at some point probably does a terminal clear and/or reset.
That tells the terminal to remove all images, but snacks is not aware of this when it happens.

folke:
It might be that kitty expect separate writes, so one thing to try would be to write each item in the queue separately instead of concatenating it

saxon1964:
I can do that, I'll get back to you soon with the results.

folke:
Just tested in kitty and the current main, always properly renders the images for me.
What Neovim and kitty version are you using?

folke:
Can you update and check again? Made some changes that may or may not fix your issue.
When opening a file from the cmdline, then in my case everything is also executed through the buffer.

saxon1964:
I have tested everything on Mac Mini M4 again, same setup, all plugins in neovim updated to the latest... The results are only marginally better:

- If the cache is empty, rendering goes perfectly well, images are visible, latex equations are rendered correctly, all of them.
- When you exit the neovim and try "nvim test.tex" again images are missing and latex equations are mostly black blocks (space is reserved properly) with just a few (say 20%) rendered correctly. 

For testing I'm using test.tex suggested in my other pull request, much more complicated than your original.

Issue #1401, opened 2025-02-23
bug(zen): `cmdline` obstructs Zen window


andreihh:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

Both NVIM v0.11.0-dev-e641155 and NVIM v0.10.4

### Operating system/version

Ubuntu 22.04.5 LTS x86_64

### Describe the bug

The Zen window height does not take `cmdheight` into account. If `cmdheight` is > 0, it will obstruct the Zen window. This is most certainly also the cause of https://github.com/folke/snacks.nvim/issues/1375:

![Image](https://github.com/user-attachments/assets/4509fb5d-715e-4a2b-a0a7-06c5bed88a94)

Setting `cmdheight` to 0 fixes the issue:

![Image](https://github.com/user-attachments/assets/a3aa83a8-e194-4738-bb8c-8a6b7d3747a3)

### Steps To Reproduce

1. Set `cmdheight` to 1 or greater (to make the bug easily noticeable, set to a larger value, e.g. 10)
2. Start Zen mode

### Expected Behavior

The Zen window should either have a smaller height to allow space for the `cmdline` (decrease window height by `cmdheight`), or should disable the `cmdline` (i.e., set `cmdheight` to 0) upon entering Zen mode and restore it upon exiting.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      "folke/snacks.nvim",
      opts = { zen = { toggles = { dim = false } } },
    },
  },
})
```

andreihh:
Prepared https://github.com/folke/snacks.nvim/pull/1402. I tested it and can confirm that it fixes both this issue and https://github.com/folke/snacks.nvim/issues/1375.

Issue #1375, opened 2025-02-22
bug(zen): scrolling with `<C-d>` in Zen mode overlays `cmdline` content on top of buffer


andreihh:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

Ubuntu 22.04.5 LTS x86_64

### Describe the bug

Scrolling down with `<C-d>` in Zen mode gets the `cmdline` contents overlaid on top of the buffer. This is a visual bug, bringing the cursor on the affected line will clear the incorrect content.

### Steps To Reproduce

1. Open Zen mode
2. Ensure the `cmdline` has non-empty contents: ![Image](https://github.com/user-attachments/assets/60b2616e-cdb6-47bd-b9a3-c2f174bfd13c)
3. Scroll down with `<C-d>`: ![Image](https://github.com/user-attachments/assets/58ffd7eb-1169-48c1-b112-693133e3cd4f)

### Expected Behavior

`cmdline` contents should not be overlaid on top of the buffer.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      "folke/snacks.nvim",
      opts = { zen = { toggles = { dim = false } } },
    },
  },
})
```

folke:
Cant reproduce

andreihh:
Can you please take another look? Make sure the `cmdline` contents are long enough to overlap the Zen window (if necessary, resize the terminal window and make it smaller).

I can reproduce on 3 different machines (1x Ubuntu, 2x Debian), 3 different terminals (xterm, Alacritty and Ghostty), and both on Neovim 0.10.4 and nightly.

PR #1402, opened 2025-02-23
fix(zen): always count cmdheight towards Zen bottom offset

andreihh:
## Description

Zen window height should always account for `cmdheight` (not only if `statusline` is enabled) and be computed dynamically when `cmdheight > 0` to avoid the `cmdline` obstructing the Zen window.

## Related Issue(s)

- Fixes https://github.com/folke/snacks.nvim/issues/1401
- Fixes https://github.com/folke/snacks.nvim/issues/1375

## Screenshots

Before: zen-mode-cmdline

After: zen-mode-fixed-cmdline


folke:
Makes sense. ty!

PR #1404, opened 2025-02-24
docs: docs/indent.md config

sohanglal:
## Description

Corrected config of docs/indent.md


folke:
euh no?
That change is totally wrong.

PR #1468, opened 2025-02-28
fix(input): don't enter `insert` mode from `normal` confirm on `input`

### LINKED ISSUE #1466: bug: closing input from normal mode switches initial buffer to input mode

### Neovim version (nvim -v)

0.10.4

### Operating system/version

Arch Linux

### Describe the bug

Hi!

After closing snacks input window by pressing "enter", initial buffer will be switched to insert mode.

For example, if I want to change variable casing, I switch normal mode and use one of the `gU<motion>` commands, then accept changes by pressing "enter" - snacks input window closes but initial buffer is switched to insert mode.

That only happens if you press "enter". In case you press "escape" - initial buffer will stay in normal mode.

### Steps To Reproduce

1. Invoke snacks input (rename a variable for example)
2. While in input window, leave insert mode (by pressing "escape" for example)
3. Can perform some normal commands, like `gU<motion>` (but that step is optional)
4. While in input window, press enter (to confirm changes and close input window)

### Expected Behavior

Initial buffer will be in normal mode

### Repro

```lua

require("lazy.minit").repro({
	spec = {
			input = { enabled = true },
		} },
		-- add any other plugins here
	},
})
```

dpetka2001:
## Description
When pressing `<CR>` for the `confirm` action on `input` prompt buffer from `normal` mode, you get dropped in `insert` mode when you get back to the normal buffer. Probably similar behavor to the `snacks.picker`.

I pondered if the `stopinsert()` should go into the `confirm` function instead, but because I also implemented `nvim_feedkeys`, so that the cursor wouldn't move to the left due to the transition from `insert` to `normal` mode, I noticed that this would move the cursor to the right when confirming from `insert` mode and thought it would be undesirable. 

That's why in the end I opted for putting this in the prompt buffer callback instead.

## Related Issue(s)
Fixes #1466

### DIFF
--- a/lua/snacks/input.lua
+++ b/lua/snacks/input.lua
@@ -233,6 +233,12 @@ function M.input(opts, on_confirm)
   vim.fn.prompt_setcallback(win.buf, function(text)
     confirm(text)
     win:close()
+    vim.schedule(function()
+      if mode ~= "i" then
+        vim.cmd("stopinsert")
+        vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<Right>", true, true, true), "n", false)
+      end
+    end)
   end)
   vim.fn.prompt_setinterrupt(win.buf, function()
     confirm()


folke:
The real issue was that the cr keymap was not bound in normal mode, so the callback kicked in, which we don't really want to use.

PR #1481, opened 2025-03-02
fix(image): markdown inline link query for shortened urls

mputz86:
## Description

  
This fixes the `image.src` capture for markdown shortcut links, which (imho) are links like `![[attachments/test.jpg]]`.


folke:
ty

Issue #1482, opened 2025-03-02
I'm on vacation until March 10th ☀️🌴🌊🏝️


folke:
I'm on vacation until March 10th.

I won't be checking GitHub and won't bring my laptop. See you all soon!

wojciech-kulik:
@folke Could you please update your plans here? It would be helpful to know when and if the development will be continued 🍻

Butterblock233:
have fun 🎉

nitaicharan:
If you’ve landed on this issue and are wondering why he hasn’t returned and snacks.nvim hasn’t received any updates, check out his comment here: https://github.com/folke/snacks.nvim/issues/1644#issuecomment-2758908856

PR #1483, opened 2025-03-02
feat: lsp_incoming_calls

sakalys:
## Description

The `lsp_incoming_calls` feature was missing

## Related Issue(s)

https://github.com/folke/trouble.nvim/issues/222

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ See the example below for how to configure `snacks.nvim`.
     { "gd", function() Snacks.picker.lsp_definitions() end, desc = "Goto Definition" },
     { "gD", function() Snacks.picker.lsp_declarations() end, desc = "Goto Declaration" },
     { "gr", function() Snacks.picker.lsp_references() end, nowait = true, desc = "References" },
+    { "gi", function() Snacks.picker.lsp_incoming_calls() end, desc = "Goto Incoming Call" },
     { "gI", function() Snacks.picker.lsp_implementations() end, desc = "Goto Implementation" },
     { "gy", function() Snacks.picker.lsp_type_definitions() end, desc = "Goto T[y]pe Definition" },
     { "<leader>ss", function() Snacks.picker.lsp_symbols() end, desc = "LSP Symbols" },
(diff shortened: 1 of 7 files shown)


dpetka2001:
Did you actually test this? There's no finder `lsp_incoming_calls`.

sakalys:
There's a caveat though. I'll add it to the original comment as well.
It overrides the internal VIM gesture:
```
							*gi*
gi			Insert text in the same position as where Insert mode
			was stopped last time in the current buffer.
			This uses the |'^| mark.  It's different from "`^i"
			when the mark is past the end of the line.
			The position is corrected for inserted/deleted lines,
			but NOT for inserted/deleted characters.
			When the |:keepjumps| command modifier is used the |'^|
			mark won't be changed.
```

sakalys:
> This isn't working for me when I checkout this PR locally. I get error `no finder lsp_incoming_calls`. And your PR also doesn't add a finder. I don't know how LSP works under the hood to give more details, but this doesn't seem to be working.

Thanks for checking. I might have confused myself somehow. Let me see what I can do address this. I will make this PR a draft. I'll have to research how to "add a finder" though, unless you can nudge me in the right direction :)

dpetka2001:
You would have to add the finder in `snacks.nvim/lua/snacks/picker/source/lsp/init.lua` I believe. But like I said I don't know LSP stuff at all, so don't know how this would work. Just that currently it's not working.

Maybe wait for maintainer's input after he comes back from vacation (see pinned #1482)

sakalys:
This is a tad bit advanced for my in understanding of lua/neovim async concepts.

It's quite simple to implement functionally, but I got stuck with nesting two async calls as one should first do a [textDocument/prepareCallHierarchy](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy), and then follow up with [callHierarchy/incomingCalls](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_incomingCalls).

Anyone care to pick this up?
This current code is missing a piece similar to the following pseudo code in [source/lsp/init.lua](https://github.com/folke/snacks.nvim/blob/63d3670d3cf6486c773c30cae51e833fe7654ea7/lua/snacks/picker/source/lsp/init.lua#L423):
````lua
function M.incoming_calls(opts, ctx)
  return function (cb)
    M.request("textDocument/prepareCallHierarchy", params, function (locationsResult)
      innerParams = locationsResult[1]

      M.request("callHierarchy/incomingCalls", innerParams, function (result)
        result.forEach(item => cb(processItem(item)))
      end
    end
  end
end

```

PR #1494, opened 2025-03-03
Update picker.md

ConnorMoodyDev:
Update the picker docs to explain usage with Oil.nvim

## Description

Adds a note to the docs to help people who want to use Oil.nvim in tander with the Snacks Picker.

### DIFF
--- a/docs/picker.md
+++ b/docs/picker.md
@@ -34,6 +34,7 @@ Some acknowledgements:
 
 The best way to get started is to copy some of the [example configs](https://github.com/folke/snacks.nvim/blob/main/docs/picker.md#-examples) below.
 
+
 ```lua
 -- Show all pickers
 Snacks.picker()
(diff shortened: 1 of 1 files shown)


ConnorMoodyDev:
Another quick note on this for @folke - if you did want it changed is adjusting the check for current_empty.

Currently in the [snacks/picker/actions.lua](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/picker/actions.lua) file on line 55/56 you have this:
```lua
local current_buf = vim.api.nvim_get_current_buf()
  local current_empty = vim.bo[current_buf].buftype == ""
    and vim.bo[current_buf].filetype == ""
    and vim.api.nvim_buf_line_count(current_buf) == 1
    and vim.api.nvim_buf_get_lines(current_buf, 0, -1, false)[1] == ""
    and vim.api.nvim_buf_get_name(current_buf) == ""
```

A proposed change to have it work in tandem with something like Oil.nvim to have it work out of the box with Oil.nvim is perhaps checking if the name of the buffer starts with oil:// 

Or alternatively, checking if the filetype is oil

PR #1511, opened 2025-03-05
fix(picker): use path in directory preview ls

milisims:
## Description

This patch lets the `directory` preview use the path determined by `Snacks.picker.util.path` to `ls` the files.

Using this in `picker.sources`:
```lua
dirs = {
  preview = 'directory',
  finder = function(opts, ctx)
    return require("snacks.picker.source.proc").proc({
      opts,
      {
        cmd = 'fdfind',
        args = { '--type', 'd', '--color', 'never', '-E', '.git' },
        ---@param item snacks.picker.finder.Item
        transform = function(item)
          item.file =  item.text
          item.cwd = opts.cwd
        end,
      },
    }, ctx)
  end
},
```


folke:
I fixed this earlier today. ty!

PR #1523, opened 2025-03-07
fix(scratch): hide buffer after formatting when close

EasonMo:
## Description

When you format code in scratch, the scratch buffer will be opened in bufferline. The buffer will not disappear even after the scratch window is closed.

## Screenshots

image
image
image


folke:
ty

PR #1534, opened 2025-03-08
fix(statuscolumn): show open folds in consecutive levels

### LINKED ISSUE #1533: bug(statuscolumn): Adjacent folds not displaying correctly with foldopen

### Neovim version (nvim -v)

0.11.0-dev-1930+gb2fa51bf15

### Operating system/version

macOS 15.3.1

### Describe the bug

When there are two adjacent folds, the part that determines whether a line is the beginning of a fold may not work properly. This is because when two folds are adjacent, the foldlevel of the previous line of the latter fold may be the same as that of the previous line, causing this line to be considered a line in a fold.

a simple example:

Image

line 361 does not display the foldopen symbol.

### Steps To Reproduce

1. Ensure the `statuscolumn` section in the configuration has the `open` option set to true
2. Create a lua file and enter the following content, making sure there are adjacent folds in the file
```lua
if true then
  print('true')
end
if false then
  print('false')
end
```
3. Problem occurs

### Expected Behavior

Correctly display foldopen symbols for adjacent folds.

### Repro

```lua
vim.o.foldlevel = 99
vim.o.foldmethod = 'expr'
vim.o.foldexpr = "v:lua.vim.treesitter.foldexpr()"

require("lazy.minit").repro({
  spec = {
    {
      "folke/snacks.nvim",
      opts = {
        statuscolumn = {
          folds = {
            open = true,
          },
        },
      },
    },

    -- install the treesitter parser for lua
    {
      "nvim-treesitter/nvim-treesitter",
      opts = {
        ensure_installed = { 'lua' },
      },
    },
  },
})

-- Example
if true then
  -- ...
end
if false then
  -- ...
end
```

dpetka2001:
## Description
This uses `vim.treesitter.foldexpr` to get the start of a foldlevel, because `vim.fn.foldlevel()` will return the same number in adjacent foldlevels, thus making the open fold markers inconsistent. 

This will however introduce a dependency on `vim.treesitter.foldexpr` and I haven't seen any other computations based on `vim.treesitter`, so maybe this is not desirable? 

Unfortunately, I wasn't able to come up with a better solution, so feel free to disregard.

## Related Issue(s)
Fixes #1533 

## Screenshots
Before 

After


dpetka2001:
I pushed a change where I check explicitly if `foldexpr` contains `vim%.lsp` or not and use the appropriate `foldexpr`. I believe this won't work correctly if the `foldexpr` that sets `vim.lsp.foldexpr()` is actually a function that gets required like in LazyVim.

folke:
## Changes Applied

I've made some improvements to this PR:

### 1. Switched to FFI-based fold detection
Instead of parsing `foldexpr` strings to detect LSP vs treesitter folds, the code now uses Neovim's C API directly via FFI:

```lua
local info = fold_info(win, lnum)
if info and info.level > 0 and info.start == lnum then
  signs[#signs + 1] = { text = vim.opt.fillchars:get().foldopen or "", type = "fold" }
end
```

**Benefits:**
- ✅ **More reliable** - Works with any fold method (treesitter, LSP, manual, expr, etc.)
- ✅ **Better performance** - Direct C API calls instead of evaluating fold expressions
- ✅ **Cleaner code** - No more hacky string parsing with `foldexpr:find("vim%.lsp")`
- ✅ **More maintainable** - Fold detection logic is unified and straightforward

### 2. Cleaned up FFI definitions
Removed unused C declarations to keep the code lean.

---

folke:
ty!

Issue #1538, opened 2025-03-09
bug: type field is not included in items send to qflist / loclist


S1M0N38:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.3

### Operating system/version

MacOS 15.3

### Describe the bug

The diagnostics items sent to qflist does not included type as expected from built-in Neovim diagnostics items.

### Steps To Reproduce

1. Open diagnostics picker (`<leader>sd` on LazyVim)
2. Send diagnostics to qflist (`<C-Q>`)
3. Inspect the qflist

Here is an example screenshot

<img width="1232" alt="Image" src="https://github.com/user-attachments/assets/7d126508-60dd-43fd-be17-dfcdc744e9ca" />

### Expected Behavior

Here is the exapected behavior screenshot

<img width="1259" alt="Image" src="https://github.com/user-attachments/assets/a2040c0a-c4a1-4d2d-b81d-7306e3d23255" />

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "folke/snacks.nvim", opts = {} },
    -- add any other plugins here
  },
})
```

PR #1539, opened 2025-03-09
fix(picker): add type field to qflist item (#1538)

S1M0N38:
Fix for #1538

## Screenshots

**Old**

Screenshot 2025-03-09 at 10 40 24

---

**After fix**
Screenshot 2025-03-09 at 10 40 58


folke:
ty!

PR #1544, opened 2025-03-09
Add pickaxe support to git

r-darwish:
## Description
Add support for filtering git log by changes (AKA pickaxe)
## Related Issue(s)

Not that I know of

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -270,6 +270,7 @@ M.git_grep = {
 ---@field current_file? boolean show current file log
 ---@field current_line? boolean show current line log
 ---@field author? string filter commits by author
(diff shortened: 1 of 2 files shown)


jakubbortlik:
Hi @r-darwish, this is a nice addition. It would be even nicer if the UI would show that the pickaxe feature is being used, something like `Git Log -S` in the input title instead of just `Git Log`.

Even better it would be if the "pickaxed" string could be modified, similar to when you can use `<c-g>` to toggle live search and narrow down your search with the `grep` and other sources.

jakubbortlik:
I'd still like this PR (or a similar implementation) to be merged.

folke:
In the git log picker, you can now switch to live mode `<c-g>` which will search using `-S`

PR #1560, opened 2025-03-12
fix(image): correct off by one issue in render fallback

### LINKED ISSUE #1557: bug: Snacks image placement off by one in floating window when using Bufferline.nvim

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

Arch Linux 6.13.5

### Describe the bug

I've actually tracked it down to the following code in Snacks.image.placement.render_fallback, I'm just not certain of the correct/best solution:

384    local pos = vim.api.nvim_win_get_position(win)
385    terminal.set_cursor({ pos[1] + 1 + border.top, pos[2] + border.left })

When Bufferline Nvim is enabled, this results in a blank line above the image and the bottom of the image running into the bottom of the floating window frame (i.e. changing it to "pos[1] + 0 + border.top..." corrects the behavior with Bufferline enabled, but then breaks it otherwise).

Image

### Steps To Reproduce

I noticed this when running under WezTerm as "placeholders"  is false, but can reproduce under kitty by modifying terminal.lua to make it use "render_fallback". Disabling bufferline.nvim also corrects the output, so it's a required factor.

### Expected Behavior

Image output should render inside the floating window box regardless of bufferline.nvim being enabled or not.

todmorrison:
Fix off-by-one issue in render_fallback when a tabline is shown (e.g. with plugins like bufferline.nvim).

## Description

When a tabline is used/shown, the positioning in render_fallback is off by one. This change checks whether a tabline would be shown and uses the correct math for that case and, otherwise, works as before.

## Related Issue(s)

  - Fixes #1557

### DIFF
--- a/lua/snacks/image/placement.lua
+++ b/lua/snacks/image/placement.lua
@@ -382,7 +382,11 @@ function M:render_fallback(state)
     self:debug("render_fallback", win)
     local border = setmetatable({ opts = vim.api.nvim_win_get_config(win) }, { __index = Snacks.win }):border_size()
     local pos = vim.api.nvim_win_get_position(win)
-    terminal.set_cursor({ pos[1] + 1 + border.top, pos[2] + border.left })
+    if ( vim.o.showtabline == 2 ) or ( vim.o.showtabline == 1 and vim.fn.tabpagenr('$') > 1 ) then
+      terminal.set_cursor({ pos[1] + border.top, pos[2] + border.left })
+    else
+      terminal.set_cursor({ pos[1] + 1 + border.top, pos[2] + border.left })
+    end
     terminal.request({
       a = "p",
       i = self.img.id,


phanen:
Maybe there're other edge case:
If I set
```
      styles = { snacks_image = { relative = 'editor', col = -1 } },
```
With this PR this image looks also wrong position.

phanen:
```diff
diff --git a/lua/snacks/image/placement.lua b/lua/snacks/image/placement.lua
index f07af64..b17f892 100644
--- a/lua/snacks/image/placement.lua
+++ b/lua/snacks/image/placement.lua
@@ -382,7 +382,10 @@ function M:render_fallback(state)
     self:debug("render_fallback", win)
     local border = setmetatable({ opts = vim.api.nvim_win_get_config(win) }, { __index = Snacks.win }):border_size()
     local pos = vim.api.nvim_win_get_position(win)
-    if ( vim.o.showtabline == 2 ) or ( vim.o.showtabline == 1 and vim.fn.tabpagenr('$') > 1 ) then
+    if
+      api.nvim_win_get_config(win).relative ~= "editor"
+      and (((vim.o.showtabline == 2) or (vim.o.showtabline == 1 and vim.fn.tabpagenr("$") > 1)))
+    then
       terminal.set_cursor({ pos[1] + border.top, pos[2] + border.left })
     else
       terminal.set_cursor({ pos[1] + 1 + border.top, pos[2] + border.left })
```

PR #1594, opened 2025-03-19
feat(picker): add toggle_regex for grep

fang2hou:
## Description

This PR adds a new keymap `<a-r>` for toggling the regex in grep picker.

It allows user to get faster when searching some special strings like `$props()`, or just put a part of line `test to jump faster.

## Screenshots

https://github.com/user-attachments/assets/2442b21c-898f-4682-a966-5a82abdd23e8

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -352,6 +352,19 @@ M.grep = {
   show_empty = true,
   live = true, -- live grep by default
   supports_live = true,
+  actions = {
+    toggle_regex = function(picker)
+      picker.opts.regex = not picker.opts.regex
+      picker:find()
+    end,
+  },
+  win = {
+    input = {
+      keys = {
+        ["<a-r>"] = { "toggle_regex", mode = { "n", "i" }, desc = "Toggle Regex" },
+      },
+    },
+  },
 }
 
 ---@type snacks.picker.grep.Config|{}


dpetka2001:
This adds this feature only to the `grep` source, but `grep_word` for example could benefit from this as well. 

Maybe something like the following could be done to generalize the toggle action?
```diff
diff --git a/lua/snacks/picker/actions.lua b/lua/snacks/picker/actions.lua                                                                                                                                           
index 6e6050e..7d5047f 100644                                                                                                                                                                                        
--- a/lua/snacks/picker/actions.lua                                                                                                                                                                                  
+++ b/lua/snacks/picker/actions.lua                                                                                                                                                                                  
@@ -633,6 +633,13 @@ function M.toggle_live(picker)                                                                                                                                                                  
   picker.input:update()                                                                                                                                                                                             
 end                                                                                                                                                                                                                 
                                                                                                                                                                                                                     
+function M.toggle_regex(picker)                                                                                                                                                                                     
+  local opts = picker.opts                                                                                                                                                                                          
+  ---@cast opts snacks.picker.grep.Config                                                                                                                                                                           
+  opts.regex = not opts.regex                                                                                                                                                                                       
+  picker:find()                                                                                                                                                                                                     
+end                                                                                                                                                                                                                 
+                                                                                                                                                                                                                    
 function M.toggle_focus(picker)                                                                                                                                                                                     
   if vim.api.nvim_get_current_win() == picker.input.win.win then                                                                                                                                                    
     picker:focus("list", { show = true })                                                                                                                                                                           
diff --git a/lua/snacks/picker/config/defaults.lua b/lua/snacks/picker/config/defaults.lua                                                                                                                           
index d5c648a..085638a 100644                                                                                                                                                                                        
--- a/lua/snacks/picker/config/defaults.lua                                                                                                                                                                          
+++ b/lua/snacks/picker/config/defaults.lua                                                                                                                                                                          
@@ -218,6 +218,7 @@ local defaults = {                                                                                                                                                                               
         ["<a-i>"] = { "toggle_ignored", mode = { "i", "n" } },                                                                                                                                                      
         ["<a-m>"] = { "toggle_maximize", mode = { "i", "n" } },                                                                                                                                                     
         ["<a-p>"] = { "toggle_preview", mode = { "i", "n" } },                                                                                                                                                      
+        ["<a-r>"] = { "toggle_regex", mode = { "i", "n" } },                                                                                                                                                        
         ["<a-w>"] = { "cycle_win", mode = { "i", "n" } },                                                                                                                                                           
         ["<c-a>"] = { "select_all", mode = { "n", "i" } },                                                                                                                                                          
         ["<c-b>"] = { "preview_scroll_up", mode = { "i", "n" } },    
```

lua/snacks/picker/config/sources.lua
```diff
@@ -352,6 +352,19 @@ M.grep = {
   show_empty = true,
   live = true, -- live grep by default
   supports_live = true,
```
jakubbortlik, line 354: Hi @fang2hou, I believe you should additionally change the value of the toggle, otherwise the `R` flag is shown when regex is *dis*abled and it's not shown when regex is *en*abled.

```suggestion
  supports_live = true,
  toggles = {
    regex = { value = true },
  },
```

See the search result with your addition (notice no `R` is shown, but the dot `.`matches a parenthesis `(`:
image

Now after toggling (disabling the default) regex the `R` is shown, but a literal `.` doesn't match the parenthesis:
image

jakubbortlik:
> This adds this feature only to the `grep` source, but `grep_word` for example could benefit from this as well.
> 
> Maybe something like the following could be done to generalize the toggle action?
> 
> ```diff
> diff --git a/lua/snacks/picker/actions.lua b/lua/snacks/picker/actions.lua                                                                                                                                           
> index 6e6050e..7d5047f 100644                                                                                                                                                                                        
> --- a/lua/snacks/picker/actions.lua                                                                                                                                                                                  
> +++ b/lua/snacks/picker/actions.lua                                                                                                                                                                                  
> @@ -633,6 +633,13 @@ function M.toggle_live(picker)                                                                                                                                                                  
>    picker.input:update()                                                                                                                                                                                             
>  end                                                                                                                                                                                                                 
>                                                                                                                                                                                                                      
> +function M.toggle_regex(picker)                                                                                                                                                                                     
> +  local opts = picker.opts                                                                                                                                                                                          
> +  ---@cast opts snacks.picker.grep.Config                                                                                                                                                                           
> +  opts.regex = not opts.regex                                                                                                                                                                                       
> +  picker:find()                                                                                                                                                                                                     
> +end                                                                                                                                                                                                                 
> +                                                                                                                                                         

This function is not necessary. The toggling actions are all created automatically in the [picker config](https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/picker/config/init.lua#L98-L102).

lua/snacks/picker/config/sources.lua
```diff
@@ -352,6 +352,19 @@ M.grep = {
...
+  actions = {
+    toggle_regex = function(picker)
+      picker.opts.regex = not picker.opts.regex
+      picker:find()
+    end,
+  },
```
jakubbortlik, line 360: This whole action is not necessary. See this [comment](https://github.com/folke/snacks.nvim/pull/1594#issuecomment-3062011703).

dpetka2001:
tnx @jakubbortlik. Wasn't aware of this :smile:

jakubbortlik:
> tnx @jakubbortlik. Wasn't aware of this 😄

Me neither, I discovered it when I was trying to add functionality that would disable toggling regex for sources that don't support it and I commented out the custom `toggle_regex` function, but the toggling keymap still was working :D

dpetka2001:
No need to use the `init` function to override the defaults. You can just put the same in the user's personal configuration plugin spec `opts.sources.grep` and it will work.

jakubbortlik:
The regex toggle only makes sense for a small number of pickers. It would be nice if the keymap would display a warning a and not toggle when it doesn't actually make senese, similar to when you press the CTRL-g mapping (toggle live) when the toggle is not supported. Otherwise you press ALT-r, an "R" is shown in the input title, but it doesn't have any effect on the behaviour, which is misleading.

dpetka2001:
Well the following check could be added to the place where the toggles actions are defined
```lua
      if name == "regex" and not picker.opts["supports_" .. name] then
        Snacks.notify.warn(
          name:sub(1, 1):upper() .. name:sub(2) .. " search is not supported for `" .. picker.title .. "`",
          { title = "Snacks Picker" }
        )
        return
      end
```
and then `supports_regex` should be added explicitly to the pickers who support it to not throw the warning. I'm guessing `grep`, `grep_word` and `grep_buffers`? Not sure if other pickers could also support it.

folke:
toggles for regex were already added, but I never added the keymap.

Issue #1599, opened 2025-03-19
bug: matcher difference from fzf


gwww:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

MacOS 15.2

### Describe the bug

I don't think I care if this is fixed. I'm reporting it because I believe that picker was trying to mimic fzf behaviour and this is a difference you may be interested in understanding.

I have a directory, with three files in it.

```
AstroLSP.lua
luasnip.lua
telescope.lua
```
Using `fzf` and searching for `lsp` yields:

<img width="148" alt="Image" src="https://github.com/user-attachments/assets/8df955d4-893b-4f91-868c-1e078964c89f" />

Using `nvim` and the snacks file picker yields a different scoring:

<img width="160" alt="Image" src="https://github.com/user-attachments/assets/ebb65250-efca-440c-b37e-0cfea3c2a762" />

Both `fzf` and snacks are run with default parameters.

When the search string is `LSP` both return the same result.

### Steps To Reproduce

See above.

### Expected Behavior

That both `fzf` and snacks return the same result.

### Repro

```lua

```

gwww:
Not stale

gwww:
Not stale

gwww:
Not stale

gwww:
Not stale

disrupted:
I am getting so tired of this bot. issues don't suddenly solve itself just because nobody has commented on it recently 🤦‍♂️

PR #1601, opened 2025-03-19
fix(picker.core): respect camelCase when ignorecase is true

gwww:
## Description

Because the search item was lowercased before it was scored camelCase would not be triggered to give the intended bonus.

This was the simplest and fastest (code execution wise) fix I could see.

## Related Issue(s)
Fixes: https://github.com/folke/snacks.nvim/issues/1599


gwww:
Not stale

gwww:
Not stale

gwww:
Not stale

gwww:
Not stale

gwww:
Not stale

folke:
ty!

PR #1608, opened 2025-03-21
feat(terminal): add position field to terminal.Opts

### LINKED ISSUE #1606: feature: Better options for position of terminal
### Did you check the docs?

To open up a floating terminal, I have to provide `cmd`, I cannot just open up a floating terminal without providing a command to run. Providing my default shell like `fish` is a workaround, but the extra delay introduced by is annoying.

I think the `snacks.terminal.Opts` should just have a position field. If not provided, the position would be determined as it used to, from `cmd`.

I tried to just pass window config with specified position to the terminal function, but it didn't seem to work. If this is doable, but I just couldn't figure it out, it should be specified in the docs.

### Additional context

I think that the default position can be configured, but I there should be an easy way of launching the terminal at whatever position you want.

karolzwolak:
## Description

I added a position field to `snacks.terminal.Opts`.
Previously, the terminal position was determined based on the `cmd`. 
If the command was provided, the position was set to "float", and it was otherwise "bottom".
The the position field is present, it overrides the position set by `cmd`, and if not provided, the old behavior is preserved.

## Related Issue(s)
- Fixes #1606

### DIFF
--- a/lua/snacks/terminal.lua
+++ b/lua/snacks/terminal.lua
@@ -22,6 +22,7 @@ local defaults = {
 
 ---@class snacks.terminal.Opts: snacks.terminal.Config
 ---@field cwd? string
+---@field position? "float"|"bottom"|"top"|"left"|"right"
 ---@field env? table<string, string>
 ---@field start_insert? boolean start insert mode when starting the terminal
 ---@field auto_insert? boolean start insert mode when entering the terminal buffer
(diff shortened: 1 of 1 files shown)


kunzaatko:
This is an option that is available using the `win` key of the options I believe. You can do this:
```lua
Snacks.terminal.open("ls", {win = {position="right"}})
```
I agree that it should be better documented though. It is not mentioned in the help and even in the README `terminal` link.

karolzwolak:
Oh yeah, I tried setting the position in win, but I must've done something wrong and it didn't work.
This should be closed in favor to improving the docs.

PR #1615, opened 2025-03-22
fix(image): set winblend=0 for floatwin when use unicode placeholders

phanen:
When `winblend > 1` (https://github.com/neovim/neovim/blob/8de1dc6923396b46c327a31daa8a1562a196a255/src/nvim/highlight.c#L751),
hover image is not shown (in kitty).

Reproduce: `nvim +'set winblend=10' docs/image.md +2`

I have a `vim.o.winblend = 20` in global config before floatwin created.
That's why I notice it.


phanen:
ping ping ping ping

folke:
ty!

PR #1616, opened 2025-03-22
fix(gitbrowse)!: make `line_end` optional when `line_count == 1`

phanen:
Add optional field (surrounded in `[...]`) and it will replaced to empty
string only when `line_count == 1`.

Not a must have feature but this can make url shorter.


phanen:
ping ping ping ping

PR #1618, opened 2025-03-23
feat(image): use a bufvar for unnamed scratch preview buffer

phanen:
I'm try integrate snacks.image into fzf-lua.

When use scratch previewer buffer, to determine the real filename a src must be explicitly given.

I don't if this way is proper but I think something should be done here..

https://github.com/ibhagwan/fzf-lua/pull/1913


folke:
I don't see why this is needed

phanen:
Sorry I should elaborate on it.

e.g. when you preview `tests/image/test.svelte` in a normal buffer, you can `vim.api.nvim_buf_get_name(buf)` to get its name.
(test.svelte use local png name, have to be resolved to actual path)


But for preview buffer, it cannot have the same name as original buffer (since it's a scratch buffer copy of the original buffer). So I use a bufvar here as the source buffer name.

PR #1629, opened 2025-03-24
fix(picker.git): add `ignore_case` for `git_grep`

### LINKED ISSUE #1627: bug: git_grep does not support ignorecase?

### Neovim version (nvim -v)

v0.10.4

### Operating system/version

Arch Linux 2025-03-23

### Describe the bug

I'm trying to figure out how to use `git_grep` picker with case insensitivity. I looked at the docs and the types of the config, and it seems like it should be doing this by default since `ignorecase = true` in default config. But my searches with this picker are not case insensitive. 

### Steps To Reproduce

1. `:cd` to git repo with a single git tracked file with the string "NewServer"
2. `:lua Snacks.picker.git_grep()` and type in "newserver" and observe lack of results
3. also try `:lua Snacks.picker.git_grep({ matcher = { ignorecase = true } })`, but this should be equivalent since that config is same as default
4. now try `:lua Snacks.picker.grep()` and typing in "newserver" and observe that it does show result

### Expected Behavior

I expect searching for "newserver" to find "NewServer" in my git repo, just as `:lua Snacks.picker.grep()` does.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
  },
})
```

dpetka2001:
## Description
Add option for `ignore_case` for `git_grep`

## Related Issue(s)
Fixes #1627

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -255,6 +255,7 @@ M.git_files = {
 ---@field untracked? boolean search in untracked files
 ---@field submodules? boolean search in submodule files
 ---@field need_search? boolean require a search pattern
+---@field ignorecase? boolean ignore case
 M.git_grep = {
   finder = "git_grep",
   format = "file",
--- a/lua/snacks/picker/source/git.lua
+++ b/lua/snacks/picker/source/git.lua
@@ -54,6 +54,9 @@ function M.grep(opts, ctx)
   elseif opts.submodules then
     table.insert(args, "--recurse-submodules")
   end
+  if opts.ignorecase then
+    table.insert(args, "-i")
+  end
   table.insert(args, ctx.filter.search)
   if not opts.cwd then
     opts.cwd = Snacks.git.get_root() or uv.cwd() or "."


dpetka2001:
I also noticed that when doing a case insensitive search even with `grep` picker the corresponding part in the preview doesn't get highlighted unless you type it case sensitive. 

If we change this line https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/picker/core/preview.lua#L326-L326
into 
```lua
local ok, re = pcall(vim.regex, "\\c" .. vim.trim(self.filter.search))
```
Then we also get highlighting for case insensitive searches. Do you see any disadvantage to not do that change? I'm not confident myself, that's why I didn't include this suggestion in the current PR.

dpetka2001:
Yes, if maintainer doesn't see any problem with this, then we could also make that change for case-insensitive highlighting in the preview window as well. I'm not confident though 100% that this won't potentially cause any other problems, that's why I just made the suggestion in the comment and wait for maintainer's input.

folke:
ty!

dpetka2001:
@folke Did you also take a look at my [comment here](https://github.com/folke/snacks.nvim/pull/1629#issuecomment-2748168988) about case insensitive preview highlight? I was waiting for your input so as to include that change in this PR as well if needed.

If you agree with this, then maybe you could create a separate commit about passing `\\c` to that call?

folke:
Is this still a problem? If so how to reproduce?

dpetka2001:
You can just do `:lua Snacks.picker.grep({ ignorecase = true })` and do a search. In the preview it should only highlight the word if it matches the case exactly. I searched for `lazyvim` and it only highlighted `lazyvim` and not `LazyVim` for example although it shows results for both.

It's for the `grep` picker, not `git_grep` that this PR is about (haven't tested with `git_grep` to be honest).

folke:
Ok feel free to make a pr

PR #1652, opened 2025-03-29
feat(explorer): add `open_all`

Sondreta:
Add a way to open all directories in the explorer at once, as a counterpart to the existing `close_all`.

This makes the explorer way more useful to me, making it possible to get an overview of the files in a project without having to press too many keys.

This implementation doesn't work recursively, which I guess can be viewed as either a bug or a feature.

## Description

See commit message. Not sure if this is something that other people want, just a suggestion. 

Thank you Folke and all other contributors to this wonderful project :)

### DIFF
--- a/lua/snacks/explorer/actions.lua
+++ b/lua/snacks/explorer/actions.lua
@@ -126,6 +126,11 @@ function M.actions.explorer_close_all(picker)
   M.update(picker, { refresh = true })
 end
 
+function M.actions.explorer_open_all(picker)
+  Tree:open_all(picker:cwd())
+  M.update(picker, { refresh = true })
+end
+
 function M.actions.explorer_git_next(picker, item)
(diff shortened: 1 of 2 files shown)


edygar:
For anyone interested on having this, you can create the action:

```lua
function(picker)
  local Tree = require "snacks.explorer.tree"
  local start = Tree:find(picker:dir())
  if start == nil then return end
  local toggle_to = not start.open

  local toggle_all
  toggle_all = function(path)
    Tree:walk(Tree:find(path), function(node)
      if node.dir then
        node.open = toggle_to
        if toggle_to then
          Tree:expand(node)
        else
          Tree:close(node.path)
        end
      end
    end, { all = true })
  end

  toggle_all(picker:dir())
  require("snacks.explorer.actions").update(picker, { refresh = true })
end
```

folke:

PR #1654, opened 2025-03-29
feat(explorer): Support Expander Icons in the Explorer Tree

Alaa-MK:
## Description
This PR supports expander icons in the explorer tree. This is native to [Neo-tree](https://github.com/nvim-neo-tree/neo-tree.nvim) and some people (including myself) are probably interested in using the snacks file explorer while having this nice directory representation. 
> [!NOTE]
> I'm not entirely sure if this should be enabled by default, so I left it disabled to be safe. This makes the feature backward compatible and only supported if users want to opt in. I think it could be a nice default though. 

## Related Issue(s)
N/A

### With icons enabled
image

### With icons disabled
image

### DIFF
--- a/doc/snacks-picker.txt
+++ b/doc/snacks-picker.txt
@@ -505,6 +505,11 @@ below.
           dir_open = "󰝰 ",
           file = "󰈔 "
         },
+        expanders = {
+          enabled = false,
+          dir = " ",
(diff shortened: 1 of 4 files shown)


Alaa-MK:
Any updates on this?

folke:

Issue #1679, opened 2025-04-03
bug(explorer): when deleteing folders it will crash


lazydiv:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

windows 11

### Describe the bug

When deleting folders in the Neovim explorer, Neovim crashes unexpectedly.  

### Steps To Reproduce

1. Open Neovim and navigate to the file explorer
2. Attempt to delete a folder using the `d` .  
3. Neovim crashes immediately after deletion.  

### Expected Behavior

**Expected Behavior:**  
The folder should be removed without causing a crash.  

**Actual Behavior:**  
Neovim crashes upon deleting the folder.  

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "folke/snacks.nvim", opts = {} },
    -- add any other plugins here
  },
})
```

PR #1684, opened 2025-04-04
fix: fixing explorer crashes when deleting

lazydiv:
Fixes a crash that occurred when deleting folders via the explorer. The issue was caused by a nil file value being passed to the file system watcher's callback, which led to a failed string concatenation. This PR adds a nil check in watch.lua to safely handle such cases.

Additionally, the explorer_del action now filters out invalid paths and prevents accidental deletion of the current working directory, improving overall stability and user safety.

Related Issue(s)
Fixes https://github.com/folke/snacks.nvim/issues/1679

Root cause: file can be nil in uv.new_fs_event callback when a directory is removed.

Screenshot 2025-04-04 104938

### DIFF
--- a/lua/snacks/explorer/actions.lua
+++ b/lua/snacks/explorer/actions.lua
@@ -257,16 +257,34 @@ function M.actions.explorer_del(picker)
   if #paths == 0 then
     return
   end
+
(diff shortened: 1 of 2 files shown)


folke:
already fixed

PR #1690, opened 2025-04-05
feat(picker): use `ivy_split` preset for sources.lines

drowning-cat:
## Description

As far as I can tell, it should be `'ivy_split'`. The `preview = "main"` property can also be deleted.


drowning-cat:
My bad.

PR #1734, opened 2025-04-11
fix(words,zen): migrate to client:supports_method for deprecation

MuntasirSZN:
## Description

Fixes deprecations by neovim 0.11.

### DIFF
--- a/lua/snacks/rename.lua
+++ b/lua/snacks/rename.lua
@@ -90,7 +90,7 @@ function M.on_rename_file(from, to, rename)
 
   local clients = (vim.lsp.get_clients or vim.lsp.get_active_clients)()
   for _, client in ipairs(clients) do
-    if client.supports_method("workspace/willRenameFiles") then
+    if client:supports_method("workspace/willRenameFiles") then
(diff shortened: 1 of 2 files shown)


MuntasirSZN:
Nope/

dpetka2001:
This will cause problems in Neovim <0.11. I don't know if Folke wants to drop support for that, but it would be better to add conditional which checks Neovim version and uses one accordingly. The old mehod is deprecated, but still resolved internally in Neovim until its removal on 0.13. It only shows the deprecated message.

MuntasirSZN:
I can write a wrapper function that handles it, but then we need to wrap client with that function, which will become ugly. @dpetka2001

dpetka2001:
I don't know how Folke wants to handle it with regards to dropping support on Neovim <0.11. That is his decision. All I know is this is not going to work on Neovim <0.11, whereas if you leave it as is both versions will work and only the deprecation message will show or a wrapper. 

Best to wait for his feedback.

MuntasirSZN:
Nope

MuntasirSZN:
Nope

MuntasirSZN:
Nope, just @folke not around.

PR #1736, opened 2025-04-11
Allow git_show previewer to use default previewer args

pedropombeiro:
## Description

Currently, I'm able to perform a search on a YADM repo by passing the args `{ "--git-dir", "/Users/pedro/.local/share/yadm/repo.git", "--work-tree", "/Users/pedro" }` to `picker.previewers.git.args`. Unfortunately, the `git_show` previewer is currently not using this configuration, so constantly fails to show the preview with an error message. This MR fixes the problem by leveraging the existing `git` local function.

## Screenshots

image

Previewer working inside a YADM repo.


folke:
ty!

PR #1743, opened 2025-04-15
fix(image): prefer wget over curl

PeterCardenas:
## Description

wget seems to work more out of the box, as with curl I get errors like:
```
curl: (3) URL rejected: Malformed input to a URL function
```

## Screenshots

before:

after:


folke:

PR #1744, opened 2025-04-15
feat(image): automatically pull git-lfs files

PeterCardenas:
Currently, when snacks image plugin encounters a git-lfs pointer to an
image, the plugin fails at the `identify` step. This is because the
plugin is trying to identify the file while it just has the pointer
metadata as contents. This commit aims to do the fix this in the
following ways:

- Adds an `lfs` step that will resolve just the pointer for the image
- Only calls this step if the image given is a git-lfs pointer
- Does some error handling in the case that git lfs fails silently


folke:

PR #1752, opened 2025-04-16
add copy_url to gitbrowse

barnabasJ:
## Description

Adds the `copy_url` function to gitbrowse to copy the url directly to the clipboard instead of opening it in the browser

### DIFF
--- a/lua/snacks/gitbrowse.lua
+++ b/lua/snacks/gitbrowse.lua
@@ -25,6 +25,11 @@ local defaults = {
     end
     vim.ui.open(url)
   end,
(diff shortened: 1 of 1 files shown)


dpetka2001:
The `open` handler for opening urls is exposed in the settings already. You can just do 
```lua
map({"n", "x" }, "<leader>gY", function()
  Snacks.gitbrowse({ open = function(url) vim.fn.setreg("+", url) end, notify = false })
end, { desc = "Git Browse (copy)" })
```
to implement this functionality which is a lot simpler.

PR #1764, opened 2025-04-19
disable mini.icons when not loaded

yaowjit:
## Description

use `MiniIcons` instead of `require("mini.icons")`, when not load `mini.icons`, there's no `MiniIcons` global variable

then, if not `require("mini.icons").setup()`, prefer to use `nvim-wbe-devicons`

## Related Issue(s)

https://github.com/folke/snacks.nvim/issues/558


folke:
no, since a lot of people lazy-load mini.icons

PR #1784, opened 2025-04-22
feat(picker): adds `center` field to `snacks.picker.jump.Config`

bnew10:
## Description

This adds the field, `center`, to `snacks.picker.jump.Config`.

Right now the default, hard-coded behavior is to `zz` after moving the cursor: https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/picker/actions.lua#L129

So that the user can configure whether the window gets centered on the line with the item that was jumped to, I added the aforementioned new field, `center`. My personal reason for adding this is I prefer the window not to scroll when the target item being jumped to is within the visible/edit window to help keep my bearings within the code (e.g., I do a `gd` on a symbol for a call to `lsp_definitions`, and the definition is already within view — and `auto_confirm` is set to true).

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -185,6 +185,7 @@ local defaults = {
     reuse_win = false, -- reuse an existing window if the buffer is already open
     close = true, -- close the picker when jumping/editing to a location (defaults to true)
     match = false, -- jump to the first match position. (useful for `lines`)
+    center = true, -- center cursorline after jump
   },
   toggles = {
     follow = "f",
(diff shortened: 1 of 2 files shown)


lua/snacks/picker/config/defaults.lua
```diff
@@ -185,7 +185,7 @@ local defaults = {
     reuse_win = false, -- reuse an existing window if the buffer is already open
     close = true, -- close the picker when jumping/editing to a location (defaults to true)
     match = false, -- jump to the first match position. (useful for `lines`)
-    center = true, -- center cursorline after jump
+    post_cmd = false, -- command to run after jump
```
bnew10, line 188: I like this better actually (`post_cmd` instead of `center`). It allows for broader behavior following the cursor jump (e.g., maybe you want it to run `norm! ztzv` to have the new cursor location at the top of the window).

folke:

PR #1799, opened 2025-04-24
Fix: Resolve glitchy scrolling during rapid keypresses in snacks.nvim

### LINKED ISSUE #1620: bug: snacks scroll has janky animations

### Neovim version (nvim -v)

0.10.4

### Operating system/version

Arch Linux

### Describe the bug

whenever I hold `j` or `k` key the animation is so janky, and it doesn't even change many lines

### Steps To Reproduce

1. install snacks and enable scroll
2. hold `j` or `k` to go down or up

### Expected Behavior

It goes up or down very smoothly

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
IDK how I can add it in there but `vim.opt.scrolloff = 21`
```

nathan-the-coder:
### Problem

When holding down the j or k keys for scrolling in Neovim with `snacks.nvim` scroll enabled, the animation becomes glitchy and slow. This issue occurs because rapid keypresses trigger multiple overlapping animations, leading to poor performance.
### Root Cause

The scroll logic did not debounce keypress events, causing unnecessary and overlapping animation calculations. This resulted in visual glitches and degraded scrolling performance.

### Fix

1. Debounce Mechanism:
    - A debounce mechanism was introduced with a default debounce time of 200ms to limit the frequency of scroll animations triggered by keypresses.
    - This ensures that rapid keypresses are handled efficiently.

2. Stop Ongoing Animations:
    - Added logic to stop any ongoing animations before starting a new one, preventing animation overlaps.

3. Configurable Debounce Time:
    - Made the debounce time configurable via the plugin's setup options, allowing users to adjust it to their preferences.

### Minimal Reproducible Setup

To reproduce the issue and verify the fix, use the following minimal Neovim configuration:
Lua

-- Save this as init.lua and run Neovim with `nvim -u init.lua`
```lua

require("lazy.minit").repro({
  spec = {
    -- Add snacks.nvim plugin
        scroll = {
          debounce_time = 200, -- Default debounce time to test the fix
        },
      },
    },
  },
  defaults = { lazy = false },
})
vim.opt.scrolloff = 21 -- Set scrolloff to enhance reproducibility
```
#### Reproducing the Issue (Before Applying the Fix)

1. Use the above configuration to start Neovim:
```bash
nvim -u init.lua
```
 2. Open any file with multiple lines of text.
 3. Hold down the j or k key to scroll up or down.
 4. Observe the scrolling behavior:
       - The animation is glitchy and feels slow.
       - The editor does not scroll smoothly due to overlapping animations.

#### Verifying the Fix (After Applying the Fix)

1. Apply the fix to the plugin's code by checking out the branch with the fix or using the modified version of snacks.nvim.
2. Use the same minimal configuration as above.
3. Open any file with multiple lines of text.
4. Hold down the j or k key to scroll up or down.
5. Observe the scrolling behavior:
    - The scrolling animation is now smooth and responsive.
    - No glitches or performance issues are observed.

#### Additional Tests
1. Adjust Debounce Time:
    - Modify the debounce_time in the configuration (e.g., set it to 50 or 300) and verify that the scrolling behavior adapts accordingly.
    - Lower debounce values may reintroduce glitches, while higher values ensure smoother scrolling.
2. Test with Mouse Scrolling:
    - Use the mouse scroll wheel to verify that scrolling animations remain smooth and unaffected by the debounce mechanism.

## Related Issue(s)
Fixes #1620

### DIFF
--- a/lua/snacks/scroll.lua
+++ b/lua/snacks/scroll.lua
@@ -29,10 +29,11 @@ local defaults = {
   },
   -- faster animation when repeating scroll after delay
   animate_repeat = {
-    delay = 100, -- delay in ms before using the repeat animation
-    duration = { step = 5, total = 50 },
+    delay = 50, -- reduced delay for faster response
+    duration = { step = 3, total = 30 }, -- shorter animation duration
     easing = "linear",
   },
+  debounce_time = 200, -- default debounce time in milliseconds
   -- what buffers to animate
   filter = function(buf)
     return vim.g.snacks_scroll ~= false and vim.b[buf].snacks_scroll ~= false and vim.bo[buf].buftype ~= "terminal"
@@ -233,6 +234,8 @@ local function scroll_lines(win, from, to)
   return vim.api.nvim_win_text_height(win, { start_row = start_row, end_row = end_row }).all + offset - 1
 end
 
+local last_scroll_time = 0
+
 --- Check if we need to animate the scroll
 ---@param win number
 ---@private
@@ -272,7 +275,18 @@ function M.check(win)
   state.target = vim.deepcopy(state.view)
   wo(win, { virtualedit = "all", scrolloff = 0 })
 
-  local now = uv.hrtime()
+  local debounce_time = config.debounce_time
+
+  -- (NEW): Added a debounce mechanism to address the scrolling issue 
+  -- where holding down the `j` or `k` keys repeatedly caused glitches.
+  -- This ensures that rapid keypresses are handled efficiently by limiting 
+  -- the frequency of scroll animations.
+  local now = uv.hrtime() / 1e6 -- convert to milliseconds
+  if now - last_scroll_time < debounce_time then
+    return -- skip if debounce time hasn't passed
+  end
+  last_scroll_time = now
+
   local repeat_delta = (now - state.last) / 1e6
   state.last = now
 


tom-gora:
With your changes applied: Holding "k" works as expected but holding "j" results in cyclical looping of the cursor range, with weird behavior where the cursor is unable to progress down past the rendered buffer section, but the range becomes smaller and smaller and at some point when it is small (just a wild hunch guess but  is it when the range traverse time falls below debounce time?) it just breaks through and scrolls down is small section jumps. Granted, with snacks as is, without this PR code, the cursor loops similarly when trying to do dirty quick scrolldown by holding "j" but seems to not be able to progress down? Nvim 0.11 (was the same on 0.10 causing me to give up on scroll) and term is Wezterm.

https://github.com/user-attachments/assets/b7035ef4-8511-43c2-9e28-088e49bff060

nathan-the-coder:
Thank's for the detailed feedback and for taking the time to test the changes! I appreciate the insights and the video demonstration.

    Regarding the behavior with the j key:
        It seems there’s still a bug causing the cyclical looping and shrinking range of motion. I’ll investigate this further to identify whether it’s related to the debounce time or the range traversal logic.
        I’ll also look into why the cursor "breaks through" after a certain point and ensure the scrolling behaves consistently.

    I’ll test the changes using Neovim 0.11 and Wezterm as you mentioned. This will help me reproduce the issue in the same environment.

    I’ll compare the current behavior of snacks.nvim (without the PR) and the behavior with the changes applied to better understand the differences and improvements.

Once I’ve identified the root cause, I’ll update the pull request with a fix. Thank you again for pointing this one out

tom-gora:
yeah I added debounce_time to  my snacks script:
```lua
	scroll = { enabled = true, debounce_time = 50 },
```
so the line
```lua 
local debounce_time = config.debounce_time
``` 
can read this at lower value but this did not change the behavior. BTW of course, I did swap the contents of `snacks.lua` where lazy installs it to your pr version but the behavior persists on my end :/

qw457812:
I care about the same issue too. I basically want to turn it off when I holding down `j` or `<C-d>` since it is too slow.

So I have set `animate_repeat.step`/`animate_repeat.total` to 0. The experience between this setting and animations disabled is not much different. Just want to let you know this.

```lua
{
  "folke/snacks.nvim",
  opts = {
    scroll = {
      animate_repeat = {
        duration = { step = 0, total = 0 }, -- holding down `j`/`<C-d>`
      },
    },
  },
}
```

dpetka2001:
I tried this locally and it still feels kind of weird although a slight improvement (even with `debounce_time = 50`).

The workaround from this [comment](https://github.com/folke/snacks.nvim/issues/1620#issuecomment-2910524926) gives me the best experience according to my own personal preferences with regards to my eyes at least.

folke:
this is just not correct and will lead to issues. closing

PR #1825, opened 2025-04-29
feat: Support rmagatti/autosession

dmtrKovalenko:
## Description

Supports this plugin for session restorations https://github.com/rmagatti/auto-session


folke:
ty

PR #1842, opened 2025-05-02
fix(win): snacks.win.Keys type (fix #1840)

### LINKED ISSUE #1840: bug: `snacks.win.Keys` type used by `snacks.win.Config` disallows table/function types

### Neovim version (nvim -v)

0.11.1

### Operating system/version

MacOS 15.4.1

### Describe the bug

`snacks.picker.win.Config` has `@field input? snacks.win.Config` ([snacks/picker/config/defaults.lua#L61](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/picker/config/defaults.lua#L61))

`snacks.win.Config` has `keys? table<string, false|string|fun(self: snacks.win)|snacks.win.Keys>` ([snacks/win.lua#L78](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/win.lua#L78))

`snacks.win.Keys` has `[1]? string` ([snacks/win.lua#L26](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/win.lua#L26))

This does not account for the shortcut syntax in [snacks/win.lua#L271](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/win.lua#L271), which allows the first item to be skipped and defined by the key in the table instead.

(which is used by [snacks/picker/config/sources.lua#L713](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/picker/config/sources.lua#L713) for example)

Presumably the `keys` field type of `snacks.win.Config` ([snacks/win.lua#L26](https://github.com/folke/snacks.nvim/blob/main/lua/snacks/win.lua#L26)) should be updated to have a new type that's equivalent to `snacks.win.Keys` with the first string index removed.

### Steps To Reproduce

```lua
require("snacks").picker.pick({ win = { input = { keys = { ["."] = { { "action1" } } } } } })
```

Observe type error
```
Cannot assign `table` to `string?`.
- `table` cannot match `string?`
- Type `table` cannot match `nil`
- Type `table` cannot match `string` [assign-type-mismatch]
```

### Expected Behavior

No type error

garrett-hopper:
See #1840 

Not sure about disabling the `cast-type-mismatch` to force the `@cast` to work on `vim.deepcopy`. I think the alternative would be to loop over the keys in `spec` and only copy over those that are not index number keys. As it is, I think it's fine since the type is corrected immediately after when the index keys are modified.

### DIFF
--- a/lua/snacks/win.lua
+++ b/lua/snacks/win.lua
@@ -276,11 +276,13 @@ function M.new(opts)
         key = { binding, spec, desc = spec }
       elseif type(spec) == "function" then
         key = { binding, spec }
-      elseif type(spec) == "table" and spec[1] and not spec[2] then
+      elseif type(spec) == "table" and spec[1] then
         --- @diagnostic disable-next-line: cast-type-mismatch allow casting as Key despite mismatched index types
         --- @cast spec snacks.win.Key
         key = vim.deepcopy(spec) -- deepcopy just in case
-        key[1], key[2] = binding, spec[1] -- replace index keys to match Key type
+        if not spec[2] then
+          key[1], key[2] = binding, spec[1] -- replace index keys to match Key type
+        end
       end
       local lhs = Snacks.util.normkey(key[1] or "")
       local mode = type(key.mode) == "table" and key.mode or { key.mode or "n" }


garrett-hopper:
Not stale

garrett-hopper:
Not stale

PR #1846, opened 2025-05-03
feat(picker): Add `default_args` file picker option

garrett-hopper:
## Description

I've added a `default_args` option to the file picker options with a default of `true`. This allows `default_args = false` to be passed along with a custom `args` table which passes all the necessary flags.

I've found a few uses where I wanted to use the builtin file picker rather than writing a custom `finder`, however I was limited by the need to use the builtin args. In particular the `-t f` assumes I'm always looking for files (naturally), but it makes it difficult to override with `-t d` to find directories only for example. (`fd` treats multiple `-t` flags as OR instead of overriding.)

Perhaps an alternative would be a flag that's more specific to the `-t`/`--type` flag. (Defaulting to `{ "f", "l" }` but being overridable with a string or a table of strings) This would potentially make more sense if it was very important that some of the default args always be passed to avoid issues. It would probably need to support `fd`, `rg`, and `find` then. 🤔

### DIFF
--- a/doc/snacks-picker.txt
+++ b/doc/snacks-picker.txt
@@ -1243,6 +1243,7 @@ FILES                                            *snacks-picker-sources-files*
     ---@field follow? boolean follow symlinks
     ---@field exclude? string[] exclude patterns
     ---@field args? string[] additional arguments
+    ---@field default_args? boolean Use default arguments (default: true)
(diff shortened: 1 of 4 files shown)


garrett-hopper:
Not stale

garrett-hopper:
Not stale

PR #1847, opened 2025-05-03
fix(scripts): Make scripts use /usr/bin/env shebang (support cross-platform)

garrett-hopper:
## Description

Tired of fixing this when running on MacOS 🙂

See https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability (`/usr/bin/env` is more universally present than `/bin/env`)

### DIFF
--- a/scripts/docs
+++ b/scripts/docs
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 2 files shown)


garrett-hopper:
Not stale

garrett-hopper:
Not stale

PR #1852, opened 2025-05-04
fix(explorer): added nil check on file for FreeBSD compatibility

joelpelaez:
## Description

NeoVim installations on FreeBSD using snacks.nvim can fail because a bug on OS file events and it can't be managed correctly by libuv.

A note about it exists in libuv documentation for [uv_fs_event_t](https://docs.libuv.org/en/v1.x/fs_event.html):

> For FreeBSD path could sometimes be NULL due to a kernel bug.

This patch will allow use snacks.nvim on FreeBSD more easily without make modifications to each installation.

## Related Issue(s)

None

## Screenshots

None


folke:
already fixed

PR #1874, opened 2025-05-09
fix(deprecated): use `client:supports_method()` instead of deprecated `client.supports_method`

willothy:
`client.supports_method` was deprecated in 0.11.


dpetka2001:
`client.supports_method` still gets resolved even in Neovim >=0.11 until its removal on Neovim 0.13. With this change there will be problems on Neovim <0.11. There should be ideally a conditional check or just leave it be as is, since the deprecation message doesn't cause any problems.

willothy:
Ah yeah good point. Not a huge deal and I see someone else has made a PR already too so I'll just close this

PR #1876, opened 2025-05-11
fix: modified words.lua to use client:supports_method instead of client.supports_method as it is deprecated

DarthMooMancer:
## Description

The client.supports_method on line 110 of words.lua is deprecated in newer versions of neovim (v0.11). I changed it to be client:supports_method as that is the new way for neovim

## Related Issue(s)

Only issue is the vim.deprecated message when loading up


dpetka2001:
See comment [here](https://github.com/folke/snacks.nvim/pull/1734#issuecomment-2871090135)

PR #1895, opened 2025-05-19
fix(images): use rsvg-convert for svg conversion

PeterCardenas:
## Description

The current `convert` method with `ImageMagick` for converting svgs to pngs is very poor and flaky. The resulting image often looks very little like the source svg or just fails to load at all. From some digging, it appears it is known that `ImageMagick` does a pretty poor job of converting svgs. I found `rsvg-convert` from [this stack overflow thread](https://superuser.com/a/723031) and it works very well.

There are some additional changes to make the all svgs use `rsvg-convert` if it is available. For fetching remote images, it seems we always call the `convert` method, even if we have an `svg` command. I put a TODO for a followup here for determining the next command after downloading a url. To fulfill this use case, I just added a conditional inside of `convert` to call `rsvg-convert`. Additionally, the `identify` cli can fail, even though it succeeds to identify the image height/width and image type. For this, I did somewhat of a hack to always continue after `identify` even if it fails, as long as the image was able to be identified.

### DIFF
--- a/lua/snacks/image/convert.lua
+++ b/lua/snacks/image/convert.lua
@@ -63,6 +63,20 @@ local commands = {
       return M.is_uri(src) and convert:tmpfile("data") or src
     end,
   },
+  svg = {
+    cmd = {
+      {
+        cwd = "{dirname}",
+        cmd = "rsvg-convert",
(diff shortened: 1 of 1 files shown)


folke:

PR #1914, opened 2025-05-25
feat(image): allow resolving src asynchronously

PeterCardenas:
## Description

Allows resolving the src of an image asynchronously. This is to support my use case of resolving images urls when previewing GitHub pull requests and issues of private repos using octo.nvim. Essentially I make some API calls to resolve the url in the PR description to the signed url that will give back the image contents without a 404 error.

## Related Issue(s)

Did not find any


folke:

PR #1918, opened 2025-05-26
fix(scope): allow user to disable keys

dpetka2001:
## Description
Currently the user is not able to disable the existing keys in `keys.textobject`, because if he sets for example `ii = false`, it throws an error about not being able to index `opts`.

## Related Issue(s)
None, rather a LazyVim discussion https://github.com/LazyVim/LazyVim/discussions/6118


folke:
ty!

PR #1920, opened 2025-05-26
fix(picker.actions): take into account if source is `recent` explicitly

dpetka2001:
## Description
I noticed in the project picker that when pressing `<c-r>` to show recent files it showed recent files from everywhere instead of only the item's directory. When the source is `recent` we have to pass `filter.cwd = ...` so that it correctly filters out recent items.

## Related Issue(s)
None


folke:
ty!

PR #1923, opened 2025-05-28
feat(notifier.dnd) option to silence all notifications

PandeCode:
Do not disturb (DND)

## Description
Sometimes something breaks and starts spamming notifications and warnings are annoying.
This toggle to turn a do not disturb feature.


folke:

PR #1925, opened 2025-05-29
feat(image): add headers to remote image requests

PeterCardenas:
## Description

Need to add some authorization headers to be able to fetch certain images, so I've added a config here to allow headers to be set for the `curl`/`wget` requests.


folke:

PR #1935, opened 2025-06-01
feat(picker - format): enables icon alignment

nickkadutskyi:
## Description
I wanted to use two icons for the directory to allow having a consistent icon for directory and to also represent an open/closed state with icons representing angle pointed left or down. E.g I'd like to have something similar to this [▷|▽] [🗂️] but this would indent folder icon further right while all other icons stay aligned left.
The goal is to allow icon alignment.

## Related Issue(s)

n/a

## Screenshots

CleanShot 2025-06-01 at 00 02 17@2x


folke:

PR #1945, opened 2025-06-05
feat(icons): allow devicons default if file icon fallback is set to an empty string

antonrogov:
## Description

When using https://github.com/nvim-tree/nvim-web-devicons and they fail to find an icon for a file, then icon from the config is used, but without any highlight, so it looks inconsistent. Devicons package has default icon and ability to customize it, so it'd be nice to have an option to use that.


folke:

PR #1946, opened 2025-06-05
fix(image): option to enable images in 256 color term

antonrogov:
## Description

Image package only uses 24-bit color to provide image id via kitty protocol. This adds support for 8-bit cterm (256 colors) by utilising foreground color of a diacritic. But since there are only 8-bit available for image ids this way, it's better to use the thrid diacritic to extend the id range by another 8-bits.

https://sw.kovidgoyal.net/kitty/graphics-protocol/#unicode-placeholders:
> By using only the foreground color for image ID you are limited to either 8-bit IDs in 256 color mode or 24-bit IDs in true color mode. Since IDs are in a global namespace there can easily be collisions. If you need more bits for the image ID, you can specify the most significant byte via a third diacritic. For example, this is the placeholder for the image ID 33554474 = 42 + (2 << 24):
> 
> ```
> printf "\e[38;5;42m\U10EEEE\U0305\U0305\U030E\U10EEEE\U0305\U030D\U030E\n"
> printf "\e[38;5;42m\U10EEEE\U030D\U0305\U030E\U10EEEE\U030D\U030D\U030E\n"
> ```
>
> Here, U+30E is the diacritic corresponding to the number 2.

Tested in Ghostty.

### DIFF
--- a/lua/snacks/image/image.lua
+++ b/lua/snacks/image/image.lua
@@ -10,7 +10,6 @@
 local M = {}
 M.__index = M
 
-local NVIM_ID_BITS = 10
 local CHUNK_SIZE = 4096
 local MAX_FSIZE = 200 * 1024 * 1024 -- 200MB
 local _id = 30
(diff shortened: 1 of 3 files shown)


folke:

PR #1956, opened 2025-06-08
feat(picker): optionally disable changing winopts in preview windows

antonrogov:
## Description

When winopts like line numbers are turned off locally (like in autocmds for terminal buffers), and those buffers are then shown in a preview window, those settings are overridden with global values (in resetting autocmd).


folke:
This is no longer relevant

PR #1959, opened 2025-06-09
fix(image): better resizing to keep ratio of original image

yuukibarns:
Problem:  The current implementation of normalize the pixel size to cell
          size applies `math.ceil()` to both `height` and `width` but if
          `height` is a little larger than 1 then it will be converted to
          2 and usually `width` is much larger than `height`, so `width`
          is roughly unchnaged, resulting in ratio x2, which makes the
          image quite narrow.

Solution: Apply `math.floor(x+0.5)` to `height` first and then calculate
          normed width by multiply the normed height with the original
          ratio.

TODO:     Since the dimensions of terminal cells varies, I think we can
          expose a threshold option for better normalization control.

### DIFF
--- a/lua/snacks/image/placement.lua
+++ b/lua/snacks/image/placement.lua
@@ -442,7 +442,7 @@ function M:state()
 
   -- scale down to fit inline
   if size.height <= 2 and is_inline() then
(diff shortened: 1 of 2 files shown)


folke:

PR #1964, opened 2025-06-11
feat(picker.git): allow passing extra args to git log command for file renames

hungpt-a8e:
## Description

Allows passing extra Git arguments (like `--git-dir` or `--work-tree`) to the rename detection logic inside `M.log` in `picker/source/git.lua`.

Previously, these extra arguments from `opts.args` were applied to the main `git log` command but ignored during rename detection. This caused errors or incorrect results when working in non-standard repo setups.

This change uses `git_args(...)` to build the rename detection command, ensuring consistent argument handling throughout `M.log`.


folke:
ty

PR #1969, opened 2025-06-12
fix(bigfile): bigfile doesn't work on windows.

AnthonyK213:
## Description

`bigfile` checks whether the path and the buffer name are identical in filetype detection, but on windows, the path separator might be different. The buffer name should be normalized here.

## Related Issue

Fixes #1722


drowning-cat:
@github-actions, Turn yourself off, please.

folke:
ty

PR #1977, opened 2025-06-15
fix(picker): remove self:attach() recursive call from M:init_layout()

### LINKED ISSUE #1913: bug: neovim crashes on resizing the terminal while `snacks.picker` is active

### Neovim version (nvim -v)

NVIM v0.12.0-dev-422+g079e7d2b11

### Operating system/version

arch linux

### Describe the bug

`Snacks.picker` causes  neovim to crash , when the terminal is resized as can be seen in the video.
https://github.com/user-attachments/assets/1f939bee-94c7-455f-86ed-60e461be19fc

### Steps To Reproduce

1. Open any of the picker from `snacks.picker`.
2. Resize the terminal.

### Expected Behavior

Picker should resize itself automatically, and should not cause neovim to crash.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
	picker = {
		prompt = " ",
		layout = {
			layout = {
				box = "horizontal",
				backdrop = false,
				width = 0.8,
				height = 0.9,
				border = "none",
				{
					box = "vertical",
					{
						win = "input",
						height = 1,
						border = "rounded",
						title = "{title} {live} {flags}",
						title_pos = "center",
					},
					{
						win = "list",
						title = " Results ",
						wo = {
						},
						title_pos = "center",
						border = "solid",
						resize = true,
						keys = {
							["<M-j>"] = function (self)
								self.action(self, "list_down")
							end,
						},
					},
				},
				{
					win = "preview",
					title = "{preview:Preview}",
					width = 0.5,
					border = "rounded",
					title_pos = "center",
				},
			},
		},
	},


} },
    -- add any other plugins here
  },
})
```

100x65b1111000:
## Description

  
 Remove infinitely looping recursive `self:attach()` call from `M:init_layout()` , which was causing neovim to crash on `VimResized` event.

## Related Issue(s)

  - Fixes #1913

### DIFF
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -268,7 +268,6 @@ function M:init_layout(layout)
       backdrop = backdrop,
     },
   }))
-  self:attach()
 
   -- apply box highlight groups
   local boxwhl = Snacks.picker.highlight.winhl("SnacksPickerBox")
@@ -284,6 +283,7 @@ function M:attach()
   -- Check if we need to load another layout
   self.layout.root:on("VimResized", function()
     vim.schedule(function()
+      -- vim.print(Snacks.picker.config.layout(self.opts))
       self:set_layout(Snacks.picker.config.layout(self.opts))
     end)
   end)
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -283,7 +283,6 @@ function M:attach()
   -- Check if we need to load another layout
   self.layout.root:on("VimResized", function()
     vim.schedule(function()
-      -- vim.print(Snacks.picker.config.layout(self.opts))
       self:set_layout(Snacks.picker.config.layout(self.opts))
     end)
   end)


Kruziikrel13:
This breaks the explorer picker and creates a hidden buffer that can be focused on.

This inconveniences window switching as you'll often end up within the hidden buffer for which you then need to jump out of it again.

# Bugged (With this PR merged)

https://github.com/user-attachments/assets/91557031-f2cb-4961-99df-08801fabd9d4


# No Bug (With this PR reverted / removed)

https://github.com/user-attachments/assets/dde0def8-9879-4897-919e-e521d607da2e

100x65b1111000:
Hi, thanks for reviewing. can you please share the relevant `explorer` config which is causing this issue, cuz I can't reproduce this.

100x65b1111000:
Also, does this only happen when you open neovim with a directory i.e something like this `nvim ~/.config/nvim`??

Kruziikrel13:
*I didn't open neovim with a directory in the above examples. I just created an empty buffer... Though I don't think that has anything to do with it. Testing by opening a directory didn't resolve the issue either.

My explorer config just has icons defined in opts. As such I'm confident that isn't the cause.


Try adding these binds though to test with
```lua
local map = vim.keymap.set
map("n", "<C-h>", "<C-w>h", { desc = "Go to Left Window", remap = true })
map("n", "<C-j>", "<C-w>j", { desc = "Go to Lower Window", remap = true })
map("n", "<C-k>", "<C-w>k", { desc = "Go to Upper Window", remap = true })
map("n", "<C-l>", "<C-w>l", { desc = "Go to Right Window", remap = true })
```

100x65b1111000:
Hi, thanks for clarification. I can indeed reproduce the issue with this PR.

Also, the issue addressed by this PR does not seem to occur with the latest nvim release, so closing this PR might be the appropriate choice.

Issue #1941, opened 2025-06-02
bug: using `put` action in registers picker


jakubbortlik:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.11.1 

### Operating system/version

Manjaro Linux 25.0.3

### Describe the bug

There are three issues with the behaviour of the built-in `put` action when used with the `registers` picker:
1. The `put` action let's Nvim figure out the text type (blockwise, charwise, linewise) by passing `""` for type in `vim.api.nvim_put`, but the `value` in the picker is always a single string, even if it should be a list of strings for linewise `put`. So when pasting a number of lines from a register with the default `put` action, the pasted text contains `^@` in place of the new line characters.
2. When the `registers` picker is opened from a floating window, the `put` action doesn't put the text into the floating window but into another window underneath.
3. The `put` action is not dot-repeatable as would be a normal `"<register_name>p`.

### Steps To Reproduce

1. Use the `Repro` below
2. Open a file with some text, or just type a few lines of text
3. Copy some linewise selection to any register
4. Run `lua vim.api.nvim_open_win(vim.api.nvim_create_buf(true, true), 1, {relative= 'editor', col=100, row=10, width=50, height=30, border="single"})` to create a new floating window
5. In the floating window type `<space>r` to open the `registers` picker
6. Select the register that contains the yanked lines
7. Press `<C-CR>` to run the default `put` action
8. See that text is pasted in another buffer than the floating window and contains `^@` characters

### Expected Behavior

1. The `put` action should handle newline characters correctly, in other words, it should respect the register type (compare the `register_aware_put` action mapped to `<CR>` in `Repro` below). For this it would be practical if the picker list also contained information about the register type (similar to what the `:reg` command does) - there is an example in the `Repro` (obviously, it would make more sense to add the `type` to the item in the [registers source](https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/picker/source/vim.lua#L261) rather than in the `format` function.
2. The `put` action should paste text into the buffer from which the picker was invoked (including floating windows)

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
[...]

PR #2012, opened 2025-07-02
fix: select correct window when closing picker

jakubbortlik:
## Description

Hi @folke, thanks for the amazing collection of plugins!

This PR fixes the bug that the incorrect window is focuses after closing a picker. This bug leads for example to pasting a register content into the incorrect buffer.

## Related Issue(s)

- partially addresses #1941

### DIFF
--- a/lua/snacks/picker/core/main.lua
+++ b/lua/snacks/picker/core/main.lua
@@ -19,7 +19,6 @@ function M.new(opts)
   local self = setmetatable({}, M)
   self.opts = opts
   self.win = vim.api.nvim_get_current_win()
-  self.win = self:find()
   return self
 end
 


jakubbortlik:
This is still relevant.

Dima-369:
Thanks so much for this trivial fix! So annoying without it when one has multiple windows!

This also works on confirming the picker selection, like for the buffer or files picker, not just on closing the picker.

docmg:
If someone is getting crazy by this Bug, you can fix it like this, until this PR is merged:
image

jakubbortlik:
> If someone is getting crazy by this Bug, you can fix it like this, until this PR is merged

That's what I've got in my config as well :grin:

isturiz:
I have the same problem in buffers like oil.nvim

folke:
That just disables main editor window logic which will maybe fix your issue, but will lead to tons of other issues.

jakubbortlik:
Thanks for fixing this properly!

docmg:
The problem with the oil.nvim buffers is still present after update @folke 
How is it possible to fix this in the snacks config?

isturiz:
> The problem with the oil.nvim buffers is still present after update @folke 
> 
> How is it possible to fix this in the snacks config? 

I’m still experiencing the issue with oil.nvim

PR #2016, opened 2025-07-03
feat(picker): improve registers picker

### LINKED ISSUE #1941: bug: using `put` action in registers picker

jakubbortlik:
## Description

This PR introduces several improvements to the registers picker:
1. Two actions are introduced (`put_register` and `put_register_before`) that make use of `vim.fn.feedkeys()` which forces the `put` action to respect the register type and to be dot-repeatable.
2. It makes the item list more informative by showing the register types ("characterwise", "linewise", "blockwise", similar to the `:reg` command) and by highlighting differently named, numbered and other registers.
3. It introduces a `toggle_reg` action (and maps it the `<c-g>` key combination which is unused for the registers picker), that enables to toggle between searching in register names or register names and contents.

## Related Issue(s)

  - Fixes #1941

## Screenshots
The two screenshots show the different highlighting and register types from point 2.

Before:
image

After:
image

### DIFF
--- a/lua/snacks/picker/config/sources.lua
+++ b/lua/snacks/picker/config/sources.lua
@@ -775,15 +775,17 @@ M.registers = {
   actions = {
     put_register = function(picker, item, action)
       ---@cast action snacks.picker.yank.Action
-      picker:close()
-      if not vim.bo.modifiable then
-        Snacks.notify("Buffer is not modifiable", { level = "error" })
-        return
-      end
-      if item then
-        local command = action and action.field == "before" and "P" or "p"
-        vim.fn.feedkeys('"' .. item.reg .. command)
-      end
+      picker:norm(function()
+        if item then
+          picker:close()
+          if not vim.bo.modifiable then
+            Snacks.notify("Buffer is not modifiable", { level = "error" })
+            return
+          end
+          local command = action and action.field == "before" and "P" or "p"
+          vim.fn.feedkeys('"' .. item.reg .. command)
+        end
+      end)
     end,
     put_register_before = { action = "put_register", field = "before" },
     toggle_reg = function(picker)


jakubbortlik:
Not stale... This is still relevant.

folke:

jakubbortlik:
Thanks for considering. It can live in my config just fine.

PR #2023, opened 2025-07-04
Update words.md

moniquelive:
Suggestion on how to customize the lsp highlighting colors

## Description

Words documentation suggestion on customizing the lsp highlight colors.


moniquelive:
@folke opinions?

folke:
That's default Neovim. no need to document this

PR #2027, opened 2025-07-05
feat(explorer): add note on requirement for fd

ian-moore:
## Description

I am on MacOSX and it was not obvious why `picker` worked on my system but `explorer` did not. I received an error message about `fd` missing, so this PR adds a note for confirmation to users that `fd` is required for `explorer` to work.


jakubbortlik:
Hi @ian-moore, the title of the PR should rather be "**docs**(explorer): add note on requirement for fd".

folke:
only required for search and it shows an error when needed

PR #2032, opened 2025-07-07
fix(explorer.watch): issue-2031: handle nil files in watch.start cb

ottobackwards:
It is possible that the watch.start callback gets called without a file when watching a directory on FreeBSD.
Handle this by adding a nil check, and setting the callback file to the path.


Kruziikrel13:
This stale bot is an absolutely horrible creation

PR #2040, opened 2025-07-10
fix(picker): do not record consecutive duplicate history

jakubbortlik:
## Description

Before this fix, identical items are added to history repeatedly. This PR fixes this by only adding an item to history if it's different from the last recorded value.

This doesn't completely remove duplicate entries in the history, it only prevents consecutive identical entries.

### DIFF
--- a/lua/snacks/picker/util/history.lua
+++ b/lua/snacks/picker/util/history.lua
@@ -53,6 +53,10 @@ function M:is_current()
 end
 
 function M:record(value)
+  -- don't record value if it's identical to the last recorded value
+  if vim.deep_equal(self.kv:get(math.max(self.idx - 1, 1)), value) then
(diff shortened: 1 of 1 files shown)


jakubbortlik:
Not stale - this is still a desired improvement.

jakubbortlik:
Keep open, s'il vous plaît.

lua/snacks/picker/core/main.lua
```diff
@@ -19,7 +19,6 @@ function M.new(opts)
   local self = setmetatable({}, M)
   self.opts = opts
   self.win = vim.api.nvim_get_current_win()
-  self.win = self:find()
```
folke: cant merge this

folke:
ty

PR #2044, opened 2025-07-12
fix(picker): properly handle smartcase, grep flag highlighting in preview and list

### LINKED ISSUE #2043: bug: picker matched text highlights are not applied with smartcase

### Neovim version (nvim -v)

v0.11.2

### Operating system/version

MacOS 15.5

### Describe the bug

Image

Image

### Steps To Reproduce

1. matcher config - `smartcase=true`(default)
2. search string includes upper case with only small case characters

### Expected Behavior

1. highlight group is applied correctly

### Repro

```lua

```

mym0404:
## Summary
Fixes issue #2043 - smartcase highlighting inconsistency between picker list and preview.

image

## Problem
When using smartcase search with mixed-case patterns, the preview highlighting was inconsistent with the list highlighting:
- List used `filter.pattern` (smartcase processed)
- Preview used `filter.search` (original pattern) 
- This caused highlighting to fail when smartcase converted patterns to case-insensitive

## Solution
1. **Added utility function** `prepare_highlight_pattern()` in `picker/util/init.lua`:
   - Removes prefix grep flags like `(?i)`, `(?-i)` etc.
   - Supports flexible trimming options (`left`, `right`, `both`, `none`)
   - Optional case-insensitive regex flag (`\c`) insertion

2. **Updated preview.lua** to use the new utility for consistent case-insensitive highlighting

3. **Updated list.lua** to use the utility for regex matcher initialization

## Test plan
- [x] Test smartcase with mixed-case search patterns
- [x] Verify consistent highlighting between list and preview
- [x] Test with various grep flags in search patterns  
- [x] Test trimming functionality

## Changes
- `lua/snacks/picker/util/init.lua`: Added `prepare_highlight_pattern()` utility function
- `lua/snacks/picker/core/preview.lua`: Use utility for consistent case-insensitive highlighting  
- `lua/snacks/picker/core/list.lua`: Use utility for regex matcher initialization

Fixes #2043

### DIFF
--- a/lua/snacks/picker/source/lsp/config.lua
+++ b/lua/snacks/picker/source/lsp/config.lua
@@ -173,7 +173,7 @@ function M.preview(ctx)
     )
   end
 
-  if item.cmd then
+  if item.cmd and type(item.cmd) == "table" then
     lines[#lines + 1] = "- **cmd**: `" .. table.concat(item.cmd, " ") .. "`"
   end
 


folke:
I fixed that in the meantime

folke:
## New Approach Implemented

Thanks for working on this! I've implemented a different approach that solves the highlighting issue more comprehensively.

### The Problem
Both PRs were trying to fix highlighting inconsistencies by adding pattern preprocessing (removing grep flags, adding `\c` for case-insensitivity, etc.). However, this approach has limitations:
- **Pattern parsing is fragile** - grep patterns can be complex (regex, fixed-string, case-insensitive combinations)
- **vim.regex quirks** - trying to replicate ripgrep's exact matching behavior in Vim regex is difficult
- **Maintenance burden** - every new grep flag or pattern type needs special handling

### The New Solution
Instead of trying to parse and reinterpret patterns, I've implemented exact match position tracking using ripgrep's built-in features:

**How it works:**
1. Use `rg --replace '__snacks__${0}__snacks__'` to mark exact matches
2. Parse the positions of marked text from ripgrep's output
3. Store positions in `item.positions` array
4. Highlight using exact character indices (no pattern matching needed)

**Benefits:**
- ✅ **100% accurate** - positions come directly from ripgrep, guaranteed to match
- ✅ **Works with any pattern** - regex, fixed-string, case-insensitive, grep flags, etc.
- ✅ **No vim.regex quirks** - no pattern conversion or workarounds needed
- ✅ **Simpler code** - no complex pattern preprocessing logic
- ✅ **Consistent everywhere** - same positions used in list, preview, and file formatter

### Implementation Details
See commit: 3b54c8d

The approach:
- Added `item.positions` field to track match indices
- Modified `grep` source to parse positions from `__snacks__` markers
- Created `highlight.matches()` helper for extmark creation
- Updated list, preview, and format to use positions when available

This supersedes the need for pattern preprocessing since we have exact match positions from the source of truth (ripgrep itself).

---

folke:
Closing in favor of the position-based approach (commit 3b54c8d). Thanks for the detailed analysis and solution attempt!

mym0404:
I'm glad to help that.

PR #2047, opened 2025-07-12
fix(scroll): can not scroll in some cases

Kaiser-Yang:
When quit from search, we can not scroll immediately.  To solve this, I use vim.schedule to refresh the state after leaving search. And after writing current buffer to a file, we can not scroll immediately. To solve this, I refresh the changedtick using BufWritePost event.

### DIFF
--- a/lua/snacks/scroll.lua
+++ b/lua/snacks/scroll.lua
@@ -157,7 +157,7 @@ function M.enable()
   })
 
   -- update current state on cursor move
-  vim.api.nvim_create_autocmd({ "CursorMoved", "CursorMovedI" }, {
+  vim.api.nvim_create_autocmd({ "BufWritePost", "CursorMoved", "CursorMovedI" }, {
     group = group,
     callback = vim.schedule_wrap(function(ev)
       for _, win in ipairs(vim.fn.win_findbuf(ev.buf)) do
@@ -167,6 +167,7 @@ function M.enable()
           -- local cursor = vim.api.nvim_win_get_cursor(win)
           states[win].current.lnum = view.lnum
           states[win].current.col = view.col
+          states[win].changedtick = vim.api.nvim_buf_get_changedtick(ev.buf)
           -- states[win].current.topline = view.topline
         end
       end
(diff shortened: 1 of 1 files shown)


folke:
I have no problems scrolling after search and really can't merge this PR with so many changes that can introduce a bunch of new issues

Kaiser-Yang:
Actually, this pr should only contain the first commit. However, it has been a long time to be merged. And during usage, I found other problems and fixed them. Maybe I should just list what the problems are.

Kaiser-Yang:
This https://github.com/folke/snacks.nvim/pull/2047/commits/eba33257fc3775f4c2435d0d67c7e5ccdc757519 is to fix the problems below:

1. After updating the contents and saving, then use <c-u> to do page  up it will not trigger the animation for the first time. This is because the chage tick is not updated after saving.
2. After using '/' or '?' and <cr> to start the search, then use <c-u> to do page up, it will not trigger the animation for the first time. This is because the state is set to nil. So I just use schedule to refresh the state.

Kaiser-Yang:
This https://github.com/folke/snacks.nvim/pull/2047/commits/7cb0c559dc6c5e5e0abddfcad2dae7756e7a4178 is to fix indent in large files. I found that in a large json files (200MB), the indent will cause neovim frozen when moving cursor. Therefore I just add a limit option to check the indent information.

Kaiser-Yang:
This https://github.com/folke/snacks.nvim/pull/2047/commits/bf826ddb14a45ceefbf2d6c5258cf3c4b446fc90 is to add a new option called middle to configure the middle part of the statuscolumn. I've achieve the feature by other plugin, therefore this commit may be useless.

Kaiser-Yang:
This https://github.com/folke/snacks.nvim/pull/2047/commits/45cf16e65a12aa9b4e4187640084b07f195f489e is to add a callback to do something after animation. This is a detour to fix some problems, one of which is when using lsp saga to use '[d' to go to the next diagnostic problem there will be a popup menu to show the options can be done by lsp. But when the next diagnostic problem is not in the same page, it will trigger the animation. And the animation will make the popup menu not show. In order to fix this, I add a option on_finished to call the lsp's show diagnostic problem under cursor. There may be a better way to solve this.

Kaiser-Yang:
https://github.com/folke/snacks.nvim/pull/2047/commits/95a715d2e581af2c57f080cbe4b036e8bec92fcd and https://github.com/folke/snacks.nvim/pull/2047/commits/7578cf3514f8feb28751d94460ac62c2a8d24f67 are to add an option for pickers to be able to decide whether or not be saved as last when using the resume picker. When I use pickers, I only hope that the files and grep pickers be saved by the resume picker. With this option, I can make it possible that the resume picker only resume the files or grep pickers.

Kaiser-Yang:
When opening two same buffers, the expand may get the wrong value (when I am in window a, actually the expand happens in window b). But when we use the nvim_win_call, we will the value from the window
which is the window we open the pickers from, which is more reasonable.

https://github.com/folke/snacks.nvim/pull/2047/commits/7578cf3514f8feb28751d94460ac62c2a8d24f67 is used to fix this.

PR #2053, opened 2025-07-16
fix(explorer): mounted directories being detected as non-directories in Tree:expand

### LINKED ISSUE #2036: bug: cannot open folder in mounted directories

### Neovim version (nvim -v)

v0.11.2

### Operating system/version

Arch/rolling

### Describe the bug

I use many virtual machines for my work and consequently I have a shared directory with my notes between them, I recently migrated to snacks-explorer, and I can no longer open folders, they have file icons, even when debugging it is possible to see that they are directories, I tested mounting a directory on my host system and got the same problem, I tried all the solutions proposed in #814 and related, and I didn't get anything different, I still have the problem.

**note:** Before I used lazyvim's neothree and it worked fine

Image

Image

### Steps To Reproduce

1. create a dummy disk
2. mount with `sudo mount -t 9p -o trans=virtio,version=9p2000.L shared /mnt/shared`
3. create directories and access them with neovim

### Expected Behavior

display directories normally

### Repro

```lua

```

n1ckisthereu:
## Description

In certain filesystems — such as bind mounts, NFS, FUSE, or other mounted paths — uv.fs_scandir_next() may return a nil type for valid directories. This causes the snacks.nvim explorer to treat mounted folders as files, making them unexpandable in the UI.

This patch adds a fallback to resolve the correct type when fs_scandir_next() returns nil. It uses uv.fs_stat() first, and falls back to vim.fn.isdirectory() if needed.

This ensures that mounted directories are properly marked as navigable.

Code change summary:

- Inside Tree:expand, we now check t == nil
- If so, we call uv.fs_stat() to get the true type
- If fs_stat fails, we fallback to vim.fn.isdirectory
- Final dir = true logic remains intact

Impact:
- Fixes a long-standing bug in mounted directories under /mnt, FUSE, etc.
- Only runs extra checks when needed (no performance hit for regular files)
- No changes elsewhere — clean, isolated patch

## Related Issue(s)
  - Fixes #2036

## Screenshots

before:
image

after:
Screenshot_20250715_222920

### DIFF
--- a/lua/snacks/explorer/tree.lua
+++ b/lua/snacks/explorer/tree.lua
@@ -141,10 +141,7 @@ function Tree:expand(node)
     if not name then
       break
     end
-    if not t then
-      local full_path = node.path .. "/" .. name
-      t = Snacks.util.path_type(full_path) or "file"
-    end
+    t = t or Snacks.util.path_type(node.path .. "/" .. name)
     found[name] = true
     local child = self:child(node, name, t)
     child.type = t
--- a/lua/snacks/picker/preview.lua
+++ b/lua/snacks/picker/preview.lua
@@ -12,14 +12,9 @@ function M.directory(ctx)
   local name = path and vim.fn.fnamemodify(path, ":t")
   ctx.preview:set_title(ctx.item.title or name)
   local ls = {} ---@type {file:string, type:"file"|"directory"}[]
-  for file in vim.fs.dir(ctx.item.file) do
-    local full_path = vim.fs.joinpath(ctx.item.file, file)
-    local file_type = Snacks.util.path_type(full_path)
-
-    table.insert(ls, {
-      file = file,
-      type = file_type,
-    })
+  for file, t in vim.fs.dir(ctx.item.file) do
+    t = t or Snacks.util.path_type(ctx.item.file .. "/" .. file)
+    ls[#ls + 1] = { file = file, type = t }
   end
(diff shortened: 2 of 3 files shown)


n1ckisthereu:
I’ve also extended the fallback fix to the preview rendering (preview.lua), which had the same issue.

On some mounted filesystems (e.g. bind mounts or NFS), vim.fs.dir() may return a nil type for valid directories. This caused mounted folders to appear with incorrect icons and ordering in the preview pane, since they were treated as files.

The fix mirrors the one in Tree:expand — we use vim.loop.fs_stat() and fallback to vim.fn.isdirectory() when needed:

```lua
if not t then
  local full_path = ctx.item.file .. "/" .. file
  local stat = vim.loop.fs_stat(full_path)
  if stat then
    t = stat.type
  elseif vim.fn.isdirectory(full_path) == 1 then
    t = "directory"
  else
    t = "file"
  end
end
```

This makes the preview behavior consistent with the tree view and ensures accurate rendering for mounted directories.
Let me know if you’d prefer to split this into a separate PR or keep it bundled.

n1ckisthereu:
In my last commit, I add the global function to validate directories, I can change to boolean function if needed, I changed too, a function named "copy_path" who also had the same problem to identify the directories mounted using fuse/bindfs, displaying errors i when try to rename/copy files.

folke:
ty!

PR #2059, opened 2025-07-16
feat(picker.files): Add support for bfs

cryptomilk:
## Description

bfs is a variant of the UNIX find command that operates [breadth-first](https://en.wikipedia.org/wiki/Breadth-first_search) rather than [depth-first](https://en.wikipedia.org/wiki/Depth-first_search). It is otherwise compatible with many versions of find.

See also https://tavianator.com/2023/bfs_3.0.html for the details why it is faster than the other implementations.


folke:

cryptomilk:
If you don't want to support multiple options, just simplify the code and only use only find.

folke:
I don't want to support too many options, and just want to keep the code as simple as possible.

Adding every feature or PR would result in a code-base that is no longer maintainable.

I literally never heard of bfs, so I assume it's a pretty niche tool. 

Most people use fd, or find.

So sorry, but I dont want this in snacks.

PR #2068, opened 2025-07-21
docs(dashboard.md): add container new dashboard example

Wabri:
## Description

I create a new example on how to use the dashboard with containers output and a choosable third panes for git logs.

## Related Issue(s)

NaN

### DIFF
--- a/docs/dashboard.md
+++ b/docs/dashboard.md
@@ -435,6 +435,99 @@ Similar to the Vim Startify dashboard
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Wabri:
Hi @folke if you have problem with billing or finance, we can try to support in some way. If you need help wrote to gabriele.puliti AT schroedinger-hat.org (I wrote like this to avoid scrapers :) )

folke:

PR #2108, opened 2025-08-03
docs(picker): add examples for picking from different file paths

xavierchanth:
## Description

I find these different pickers useful, and they all work with just snacks + native neovim apis.
The only one that may cause issues is the LSP one if you haven't migrated to native LSP yet.

This should be an easy copy/paste example for most people to add to their config.
I can reformat it to be a lazy.nvim keys table if you'd like.

## Screenshots

image

### DIFF
--- a/docs/picker.md
+++ b/docs/picker.md
@@ -595,6 +595,35 @@ Snacks.picker.pick({source = "files", ...})
 }
 ```
 
+### Search Different File Paths
+
+```lua
+-- All of these examples fallback to using neovim's CWD
+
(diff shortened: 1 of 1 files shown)


folke:
those examples are just not good, sorry. wont add

you can alos just use `vim.fn.getcwd(0)` in your own config

xavierchanth:
> those examples are just not good, sorry. wont add
> 
> 
> 
> you can alos just use `vim.fn.getcwd(0)` in your own config

No problem, they actually each serve a valid use case for me, but I understand most people won't care.

PR #2112, opened 2025-08-04
feat(picker): add resumable support

### LINKED ISSUE #1530: feature: add `resumable` picker option
### Did you check the docs?

I'd like to migrate to snacks explorer. One of the things that is bothering me is that after opening and closing (I'm using `auto_close`), it overrides the last opened picker and `Snacks.picker.resume()` then opens the explorer.

Add an option like `resumable = true | false` to picker options, so we can filter out pickers that we don't want to override the last opened picker.

none

### Additional context

_No response_

VicSegers:
## Description

Add an option resumable = true | false to picker options, define pickers you don't want to open with `Snacks.picker.resume()`.

## Related Issue(s)

- Fixes #1530

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -71,6 +71,7 @@ local M = {}
 ---@field live? boolean when true, typing will trigger live searches
 ---@field limit? number when set, the finder will stop after finding this number of items. useful for live searches
 ---@field ui_select? boolean set `vim.ui.select` to a snacks picker
+---@field resumable? boolean when true, picker resumable with `Snacks.picker.resume()`
 --- Source definition
 ---@field items? snacks.picker.finder.Item[] items to show instead of using a finder
 ---@field format? string|snacks.picker.format|string format function or preset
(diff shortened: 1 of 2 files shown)


wvffle:
bad bot

folke:
Thank you for this PR addressing the resumable option!

I've implemented an alternative solution in commit c210439 that achieves the same goal with a more flexible approach.

## Your Use Case: Excluding Certain Pickers
Your PR added a `resumable` boolean option to prevent specific pickers (like explorer) from being saved to `M.last`.

The new implementation solves this differently by:

1. **Per-source state storage**: Each picker source maintains its own resume state keyed by source name
2. **Flexible resume API** with include/exclude options:
```lua
-- Resume only specific sources
Snacks.picker.resume("files")
Snacks.picker.resume({ include = {"files", "grep"} })

-- Exclude specific sources
Snacks.picker.resume({ exclude = {"explorer", "notifications"} })
```

This means you can:
- Resume specific pickers by name: `Snacks.picker.resume("files")`
- Resume anything except explorer: `Snacks.picker.resume({ exclude = {"explorer"} })`
- All pickers save state, but you control which ones to resume via the API

## Why This Approach
Instead of a per-picker `resumable` flag that needs to be configured upfront, the new approach:
- ✅ Gives runtime control over what to resume
- ✅ Allows resuming any past picker, not just the last one
- ✅ More flexible - same functionality without config changes

Your PR identified the real need (controlling what gets resumed), and the new implementation addresses it with a more powerful API.

Thanks for the contribution!

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #2153, opened 2025-08-25
docs(rename): netrw example

cherryramatisdev:
## Description

I recently got really into improving my experience with netrw, in one of those explorations I managed to integrate with the `Snacks.rename` module by remapping the default rename mapping. So I thought it would be cool to include a more native approach to the documentation of this module.

> PS: In the example I implemented with the `vim.ui.input` interface since it should be the new standard, but if you prefer with the original `vim.fn.input` I can change it.

## Related Issue(s)

N/a

## Screenshots

[asciicast](https://asciinema.org/a/735692)


folke:
ty

PR #2178, opened 2025-09-11
feat(picker.git): improve specifying arguments for git pickers

jakubbortlik:
## Description

This PR adds:
- the possibility to specify pathspec patterns to the `git grep .... -- <patchspec>` command. This is useful for ignoring large git-tracked files from the `git grep` search.
- the possibility to specify additional arguments to `git log <log_args>` which makes it possible to pass arguments like `--no-merges`.


lua/snacks/picker/source/git.lua
```diff
@@ -54,7 +54,27 @@ function M.grep(opts, ctx)
...
+    local function is_valid(spec)
+      if not spec then return false end
+      if spec:match("^:[!^][*.]?$") then return false end
+      if spec:match("^:%(exclude%)[*.]?$") then return false end
+      if spec:match("^:%([^)]*$") then return false end -- unfinished `(exclude)`
+      return true
+    end
+    return vim.tbl_filter(function(spec) return is_valid(spec) end, pathspecs)
+  end
+  vim.list_extend(args, filter_catch_all_exclusions(pargs))
```
jakubbortlik, line 72: This part enhances the UX when entering a excluding pathspec - without this, if the pathspec is a "catch-all" like `:!*`, all items are matched and temporarily hidden. With this improvement, a catch-all expression is removed from the pathspecs which makes it possible to copy the path the user wants to actually ignore.

folke:
I just pushed some changes to allow configuring git and git cmd args for all git sources.

I also resolved conflicts in this PR, but it has to be manually fixed I think.

Issue #1732, opened 2025-04-11
bug: race condition in git checkout via picker


pfnilsson:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.0

### Operating system/version

MacOS 15.3.2

### Describe the bug

When using the git_branches picker and checking out a branch it sometimes (more often than not in my case) errors out. I added some extra debugging because the error message did not give anything to work with - and it showed that there is a lock file:

Output:
fatal: Unable to create '/path/to/my/repo/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.

The command always works if I try to checkout on the command line and it also works from within neovim (both as a standalone command and via LazyGit).

I have not been able to find that any other process is influencing this, and I also found that it works 100% of the time if I add a delay like so:

```
function M.git_checkout(picker, item)
  picker:close()
  if item then
    local what = item.branch or item.commit --[[@as string?]]
    if not what then
      Snacks.notify.warn("No branch or commit found", { title = "Snacks Picker" })
      return
    end
    local cmd = { "git", "checkout", what }
    local remote_branch = what:match("^remotes/[^/]+/(.+)$")
    if remote_branch then
      cmd = { "git", "checkout", "-b", remote_branch, what }
    end
    if item.file then
      vim.list_extend(cmd, { "--", item.file })
    end

    -- Add a delay to account for transient locks
    vim.defer_fn(function()
      Snacks.picker.util.cmd(cmd, function(output, code)
        if code == 0 then
          Snacks.notify("Checkout " .. what, { title = "Snacks Picker" })
          vim.cmd.checktime()
        else
          Snacks.notify.error("Git checkout failed: " .. table.concat(output, "\n"))
        end
      end, { cwd = item.cwd })
    end, 500)  -- 500ms delay; adjust as needed
  end
end
```

Should also note that it always works on the 2nd try of checking out a branch - so I'm not sure if the first one is doing something extra 🤔 

### Steps To Reproduce

1. Check out a branch with the git_branches picker
2. See error

[...]

PR #2181, opened 2025-09-12
fix(words.lua): update client.supports... (deprecated) -> client:supports…

zledev:
Title: fix(words.lua): client.supports... (deprecated) -> client:support…
Description: Fixes nvim deprecation error.

This occurs on Latest version of Nvim (NVIM v0.12.0-dev)

image


dpetka2001:
Duplicate #1732. Also this will cause problems in Neovim versions <0.11. Maintainer should decide what the minimum Neovim version he wants to support should be.

zleXece:
@dpetka2001 Thank you for pointing that out. I was not aware of such duplication and the issue. Indeed you have a great point, the Maintainer should decide what the minimum version he wants to support.

folke:
Fixed in the meantime

PR #2190, opened 2025-09-17
[grok] fix(explorer/watch): handle nil filename in fs_event callback

Dmitry-N-Medvedev:
## Description

On FreeBSD (kqueue backend), the filename parameter can be nil due to a kernel
bug in LibUV's fs_event. Add a nil check to refresh the directory and avoid
concatenation errors, preserving watch functionality for LSP diagnostics
(e.g., ZLS in Zig files).

### DIFF
--- a/lua/snacks/explorer/watch.lua
+++ b/lua/snacks/explorer/watch.lua
@@ -16,6 +16,14 @@ function M.start(path, cb)
   end
   local handle = assert(vim.uv.new_fs_event())
   local ok, err = handle:start(path, {}, function(_, file, events)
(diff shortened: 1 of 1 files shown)


luisdavim:
Is this a duplicate of https://github.com/folke/snacks.nvim/pull/2032 ?

Dmitry-N-Medvedev:
> Is this a duplicate of #2032 ?

looks like it.

joelpelaez:
My patch only handled the nil filename but I don't know how explorer works, so my PR is too simple. I hope than this PR will be merged soon.

PR #2214, opened 2025-09-27
fix(dashboard): better window resize handling

farzadmf:
## Description
- If Neovim window is quickly resized after dashboard is shown, debounce the render for 200ms to avoid a second render.
- Cache terminal window buffer and reuse it on resize.

## Related Issue(s)
- Didn't create issue but can create one if needed

## Screenshots
- I don't have an "interesting" screenshot; let me know if it's required.

## Additional info
In my config, I have AutoCmds to hide lualine in the dashboard (for whatever reason, couldn't make it work through lualine's config), and that causes a resize in dashboard and a second render once I start Neovim. I've been testing out these changes for a few days, and they seem to be working.

PS: I _have_ used AI to help with changes.

PS2: Feel free to close/ignore this PR if you think is not necessary. Since the changes seem to be fine, I thought it would be nice to contribute.


folke:

PR #2215, opened 2025-09-28
fix(image): hover close in insert mode

jZhangTk:
## Description

I'm not 100% sure if this is a bug or if it was intentional, but the `return` seems to have been lost from a previous refactor https://github.com/folke/snacks.nvim/commit/e35d6cd4#diff-ff9a4f8a621887c466394f755193cc78a5e69e0e2e86f4e668662e1434d3611eL878.

When an image displays using float or hover, the image can block the text even in insert mode. This makes editing really hard. Adding the `return` can close the hover in insert mode.

## Related Issue(s)

I couldn't find an issue created for this, but there is a discussion about it https://github.com/folke/snacks.nvim/discussions/1777.

## Screenshots

Before the fix, the image shows in insert mode.
image

After the fix, the image stops showing in insert mode.
image


jZhangTk:
Besides adding the `return`, I realized we need to split up the `hover.win:valid()` check and not let it return.

When jumping between two images directly (without any other interim cursor movements), `hover.win:valid()` can result in `false`, and it will not continue to show the image if we have the `return`.

folke:
ty!

PR #2225, opened 2025-09-30
fix(indent): do not render ext marks for folded lines

### LINKED ISSUE #2222: bug: Indent chunk should not render when folded (clones #352, still present)

### Neovim version (nvim -v)

v0.11.4

### Operating system/version

MacOS 15.7

### Describe the bug

indent markers or spaces are rendered in folded lines with indent enabled.

basically a clone of closed : https://github.com/folke/snacks.nvim/issues/352

Image
Image
Image
Image

### Steps To Reproduce

1. write some nested python, ideally with 100+ lines in the fold. e.g. 
```python
for a in range(5):
    for b in range(5):
        for c in range(5):
            for d in range(5):
                for e in range(5):
                    for f in range(5):
                        for g in range(5):
                            print("ok")
                            print("ok")
```
2. `:set foldmethod=indent`

### Expected Behavior

indent does not print spaces or markers on folds.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    { "folke/snacks.nvim", opts = {indent={enabled=true}} },
  },
})
```

sfarina:
## Description

I find indent extremely helpful, but also make heavy use of cold folding, where the existing behavior distracts me. This fixes what I see as a minor bug in the indent snack.

## Related Issue(s)

  - Fixes #2222 

## Screenshots

old behavior:
Screenshot 2025-09-29 at 8 03 20 PM

new behavior:
Screenshot 2025-09-29 at 8 05 03 PM

### DIFF
--- a/lua/snacks/indent.lua
+++ b/lua/snacks/indent.lua
@@ -270,8 +270,11 @@ function M.on_win(win, buf, top, bottom)
       end
       indent = math.min(indent, parent_indent + state.shiftwidth)
       local extmarks = show_indent and indent > 0 and get_extmarks(indent, state)
-      for _, opts in ipairs(extmarks or {}) do
-        vim.api.nvim_buf_set_extmark(buf, ns, l - 1, 0, opts)
+
+      if vim.fn.foldclosed(l) == -1 then
+        for _, opts in ipairs(extmarks or {}) do
+          vim.api.nvim_buf_set_extmark(buf, ns, l - 1, 0, opts)
+        end
       end
     end
   end)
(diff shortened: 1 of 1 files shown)


folke:
this is too expensive performance wise to do this in that code path. sorry

sfarina:
that's ok, thanks for the review.

folke:
There is a simple work-around though, you can set a bg to `Folded`. Like this is how it looks for me:

image

That's also with `vim.o.foldtext = ''` to get the highlighting

PR #2239, opened 2025-10-03
fix(notifier): include icon in padding in minimal style

### LINKED ISSUE #2238: bug(notifier): minimal style doesn't account for icon when padding

### Neovim version (nvim -v)

NVIM v0.11.4

### Operating system/version

MacOS 15.6.1

### Describe the bug

When a notification is longer than min_width, it starts to flow into the space where the icon ext_mark is drawn:

Image

### Steps To Reproduce

1. Generate a long notification, e.g. `vim.notify('01234567890123456789012345678901234567')`. The actual length required will depend on `vim.o.columns`

### Expected Behavior

Padding should account for icon width when the minimal style is used:

Image”

### Repro

```lua
vim.env.LAZY_STDPATH = '.repro'
load(vim.fn.system('curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua'))()

require('lazy.minit').repro({
  spec = {
    {
      'folke/snacks.nvim',
      opts = {
        notifier = {
          enabled = true,
          style = 'minimal',
          timeout = 10000,
        },
        styles = {
          notification = {
            wo = {
              wrap = true, -- enable line wrapping in notifications
            },
          },
        },
      },
    },
    -- add any other plugins here
  },
})
```

cameronr:
## Description

This prevents the text from running underneath the icon when the text is a single line and the text plus padding is less than max_width.

It does not handle the case where the text is longer than max_width (with or without wrap).

## Related Issue(s)

Fixes #2238 

## Screenshots

Before:

Screenshot 2025-10-03 at 13 52 39

After:

Screenshot 2025-10-03 at 13 53 33

### DIFF
--- a/lua/snacks/notifier.lua
+++ b/lua/snacks/notifier.lua
@@ -581,7 +581,9 @@ function N:render(notif)
 
   local lines = vim.api.nvim_buf_get_lines(buf, 0, -1, false)
 
-  local pad = self.opts.padding and (win:add_padding() or 2) or 0
+  -- for the minimal style, we also have to factor in the icon width
+  local icon_width = self.opts.style == "minimal" and vim.api.nvim_strwidth(notif.icon) or 0
+  local pad = (self.opts.padding and (win:add_padding() or 2) or 0) + icon_width
   local width = win:border_text_width()
   for _, line in ipairs(lines) do
     width = math.max(width, vim.fn.strdisplaywidth(line) + pad)


tkapous (approved):
clean, but should maybe use `vim.api.nvim_strwidth`?

cameronr:
> clean, but should maybe use `vim.api.nvim_strwidth`?

Yeah, I think that would be a bit better. I re-tested it and will re-push.

folke:
ty

PR #2242, opened 2025-10-05
fix(scratch): use icon[1] when icon is a table to avoid table.concat error

hifanx:
## Description

Problem: opts.icon can be a table {icon, icon_hl} but the code inserts opts.icon directly into filekey and later does table.concat(filekey, "|"). That crashes when an element is a table.

Fix: use the first element of opts.icon when opts.icon is a table (fallback to opts.icon or "").

## Related Issue(s)

N/A

## Screenshots

This happens when scratch.opts.icon is a table per documentation
image


folke:
ty

PR #2244, opened 2025-10-05
feat(explorer): add lualine integration for explorer breadcrumb

cavanaug:
## Summary

Adds a lualine component that displays breadcrumb/path information for the snacks.nvim explorer in the statusline.

## Features

- New `snacks_explorer` lualine component at `lua/lualine/components/snacks_explorer.lua`
- Configurable path display formats:
  - `relative` - path relative to explorer cwd (default)
  - `display` - `~`-relative path
  - `filename` - just the filename
- Shows file/directory icons (󰈔 for files, 󰉋 for directories)
- Only displays when explorer window has focus
- Auto-updates as you navigate through the explorer

## Implementation

- Added state tracking in `lua/snacks/explorer/actions.lua` with `update_explorer_path()` and `clear_explorer_state()` functions
- Hooked into explorer via `on_change` and `on_close` callbacks in `lua/snacks/picker/config/sources.lua`
- Component integrates with lualine's component discovery system (similar to aerial.nvim)
- Includes documentation in `docs/explorer.md` with setup examples

## Usage

\`\`\`lua
{
  "nvim-lualine/lualine.nvim",
  opts = function(_, opts)
    table.insert(opts.sections.lualine_c, { "snacks_explorer", path_type = "relative" })
  end,
}
\`\`\`

### DIFF
--- a/docs/explorer.md
+++ b/docs/explorer.md
@@ -94,3 +94,43 @@ Reveals the given file/buffer or the current buffer in the explorer
 ---@param opts? {file?:string, buf?:number}
 Snacks.explorer.reveal(opts)
 ```
+
+## 🎨 Lualine Integration
+
+The explorer provides a lualine component that displays the current file/directory path in the statusline when the explorer window has focus.
+
+### Setup
+
+Add `"snacks_explorer"` to your lualine configuration:
+
+```lua
+require("lualine").setup({
+  sections = {
+    lualine_c = {
+      "filename",
+      "snacks_explorer",
+    },
+  },
+})
+```
+
+### Options
+
+Configure the path display format:
+
+```lua
+{ "snacks_explorer", path_type = "relative" }  -- relative to cwd (default)
+{ "snacks_explorer", path_type = "display" }   -- ~-relative path
+{ "snacks_explorer", path_type = "filename" }  -- just filename
+```
+
+### LazyVim Example
+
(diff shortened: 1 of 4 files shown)


cavanaug:
This may not have been the right way to do this, but I took a stab at it.   In fact it probably isnt.   Im not a lua/vim expert, I tried guiding my intern claude to do the right thing here, but Im not sure he did it properly.

The goal was simply to add breadcrumbs in the lualine for the current line in the explorer pane.  This is similar to how say editing a json file would populate the path to the current cursor location.

The reason this is useful is I frequently use the explorer in situations where the directory has multiple files/folders with a common long prefix that doesnt allow for the full file/folder name to be visible in the narrow explorer pane (think com.java.xx.x.xx.x.xxxxx).   Having the full relative path in the lualine makes it so i can differentiate quickly.

folke:
I'm not a fan of this.  Thank you anyway

cavanaug:
@folke Do you have a suggested recommendation for how to see an entire filename or directory when the explorer pane is too narrow to disambiguate them???   That’s the use case I was trying to solve for.  Or is it a “yeah that won’t work” sort of thing.

folke:
I just pushed a change where you can now do `<c-g>` (like in any nomal Neovim buffer) to print the cwd

PR #1497, opened 2025-03-04
feat(picker): keymaps: explorer_expand / explorer_collapse

tkapous:
## Description

This adds `explorer_expand` and `explorer_collapse` actions to the picker, and bind them to `>` and `<` for the explorer list.

The reasoning behind the keymap is that it's an indenting of sorts.

The actions are merely calling `Tree:open` and `Tree:close`. So, the naming is inconsistent but logical in the context. I think it's okay, but there might be a debate here.

PR #2246, opened 2025-10-05
feat(picker): keymaps: explorer_expand / explorer_collapse

tkapous:
I closed #1497 by mistake.


folke:
you can already do this using `h/l` keys.

tkapous:
Yes, mostly, but `l` opens files as well. I think it'd  make sense to have both behaviours, but that's just me. 
Thanks for the update!

PR #2250, opened 2025-10-07
fix(picker): restore LSP picker items on resume

### LINKED ISSUE #1579: bug(picker) LSP pickers fails to resume

### Neovim version (nvim -v)

0.10.4

### Operating system/version

MacOs

### Describe the bug

When trying to resume an LSP picker, the error is shown: 

> No results found for `lsp_references`

### Steps To Reproduce

1. Go to a variable, or any symbol
2. run `:Snacks.picker.lsp_references()` 
3. Select an item or close the picker
4. Move the cursor to an empty line
5. run: `Snacks.picker.resume()` 
6. The error is shown, and an empty picker is displayed

### Expected Behavior

I would expect the picker to resume and show the same results, regardless of the cursor position. 

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

carlos-algms:
## Description

- Add restore() method to finder to reuse cached items
- Cache items for LSP sources in `M.last` on close
- Restore cached items when resuming LSP pickers
- Add comprehensive tests for LSP picker resume functionality

## Related Issue(s)

- Fixes #1579

### DIFF
--- a/lua/snacks/picker/core/finder.lua
+++ b/lua/snacks/picker/core/finder.lua
@@ -79,6 +79,16 @@ function M:init(filter)
   return ret
 end
 
+---@param items snacks.picker.finder.Item[]
+---@param filter snacks.picker.Filter
+function M:restore(items, filter)
+  self:abort()
+  self.items = items
+  self.filter = filter
+  -- Mark task as complete so matcher can process items
+  self.task = Async.nop()
+end
+
 ---@generic T: table
 ---@param t T
 ---@return T
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -44,6 +44,7 @@ M._active = {}
 ---@field opts? snacks.picker.Config
 ---@field selected snacks.picker.Item[]
 ---@field filter snacks.picker.Filter
+---@field items? snacks.picker.Item[]
 
 ---@type snacks.picker.Last?
 M.last = nil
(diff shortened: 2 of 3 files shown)


carlos-algms:
Thanks for adding the label @folke, however, it could be misleading. 
The actual "code" changes are low, less than 50 lines, including luaDocs, and everything else is unit tests.

folke:
The label is automatic

folke:
Thank you for this excellent PR! I really appreciate the thorough implementation with tests.

I've implemented a similar solution in commit c210439 that refactors the entire resume functionality. The new implementation includes all the features from your PR:

## LSP Caching (Your Key Feature)
Your PR's main innovation - caching LSP items to resume from anywhere - is now implemented in `resume.lua:21`:
```lua
items = source:find("^lsp_") and picker.finder.items or nil
```

When resuming, cached items are restored via a custom finder (resume.lua:32-34):
```lua
if state.items then
  state.opts.finder = function()
    return state.items
  end
end
```

This achieves the same goal as your `finder:restore()` method but without needing to modify the Finder class.

## Additional Benefits
The new implementation goes further by:
- **Per-source state storage**: Each picker source maintains its own resume state independently (not just one `M.last`)
- **Flexible resume API**: `Snacks.picker.resume("lsp_references")` or with include/exclude options
- **Timestamp-based selection**: Automatically picks the most recent picker when multiple match
- **Cleaner separation**: Resume logic moved to dedicated `picker/resume.lua` module

## Why Close This PR
While your implementation is solid and well-tested, the new refactored approach provides a superset of functionality without requiring changes to the Finder class. The LSP caching feature you pioneered is now part of the core resume functionality.

Your work definitely helped validate this approach - thank you for the contribution!

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

carlos-algms:
Thank you @folke I just updated it using Lazy, and I can confirm the resume works for the LSP picker. 

👏🏻 🎉

PR #2253, opened 2025-10-09
feat(picker.projects): make max_depth customizable

joelazar:
## Description

By making the `max_depth` customizable, I could provide a single entry dir to the projects picker to discover all my repos 🤞


folke:
ty

Issue #1548, opened 2025-03-10
bug: lazygit config creation blocked


BirdeeHub:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.11.0-nightly+e02ee74

### Operating system/version

NixOS 25.05 (linux, any)

### Describe the bug

If you delete the config file for lazyvim, open neovim and try to open it via snacks, it errors out.

```
2025/03/08 23:49:31 stat /home/t-elos/.config/lazygit/config.yml: no such file or directory

[Process exited 1]
```

Its not enough to just create `~/.config` or even `~/.config/lazygit`

You have to create the actual `~/.config/lazygit/config.yml` file or it errors out

Lazygit itself on its own is able to handler this, so it shouldn't be happening.

### Steps To Reproduce

Install snacks.nvim and lazygit

Delete ~/.config/lazygit/config.yml

Open lazyvim via snacks

Observe error

### Expected Behavior

It should make the file if it doesnt exist, or prevent 

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "folke/snacks.nvim", opts = {
      lazygit = {},
    } },
  },
})
```

BirdeeHub:
https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/lazygit.lua#L91-L93

It looks to me like here you should be checking if the file exists first before adding it.

dpetka2001:
In my personal opinion I don't believe it's Snacks' responsibility to ensure that the user has a working lazygit configuration. 

The assumption is that the user already has a working lazygit configuration and `snacks.lazygit` only adds to that. That's why the error is also thrown. So, the user should just run lazygit once outside of Neovim for the config file to be populated and then run `snacks.lazygit`.

What if the user doesn't have lazygit installed? Should Snacks install it for the user?

BirdeeHub:
Thats exactly the PR I was planning on doing I just hadnt gotten around to it yet XD

dpetka2001:
[...]

PR #2256, opened 2025-10-10
fix(lazygit): check if default config file exists before adding to LG_CONFIG_FILE

cosmicbuffalo:
## Description

I noticed that lazygit was erroring out on open when the default config file was missing even though lazygit outside of neovim opened fine without it. Tracked it to the `LG_CONFIG_FILE` var that tells lazygit to error out when any of the files are missing - and snacks was setting this to include the default config even when it didn't exist. 

## Related Issue(s)

fixes issue: https://github.com/folke/snacks.nvim/issues/1548


folke:
ty

PR #2266, opened 2025-10-13
feat(picker.lsp_symbols): add "flatten" option to lsp_symbols

DanWlker:
## Description

lsp_symbols display keys in a tree with only the current key of the symbol displayed. It does not allow searching for keys in their full length ex. if the key is`a.b.c.d`

This is useful in languages like json, yaml and toml.

### Before:
image

When searching:
image

### After:
image

When searching
image


folke:
Not interested in adding this.

PR #2269, opened 2025-10-14
fix(image): skip `\usepackage` not at the start of a line

danielwe:
## Description

Currently, the `snacks.image` LaTeX preview may fail if the document uses `\usepackage` inside another macro. For example, consider the following:

```tex
\IfFileExists{mypackage.sty}{\usepackage{mypackage}}{}
```

The context in this particular example is that `mypackage.sty` is a local file defining a package. LaTeX will error out if I do `\usepackage{mypackage}` and the file doesn't exist, so I guard it with `\IfFileExists` for robustness.

However, when parsing this line, `snacks.image`s logic ends up adding `mypackage.sty` to the list of package names, because it's the first macro argument on a line that contains `\usepackage`. This obviously fails; however, improving the parsing to add `mypackage` instead of `mypackage.sty` wouldn't help, as my local package doesn't exist in the directory where the `snacks.image` preview is compiled, so LaTeX would still error out.

This PR proposes the simplest possible workaround: only consider `\usepackage` statements at the start of a line. This is obviously a bit of a hack, based on the simple heuristic that if `\usepackage` is inside some other macro, there's probably some funny business going on and it's probably best to ignore it for the `snacks.image`s purposes. It's not foolproof, but at least it enables one way of using local packages without breaking the preview.

Let me know if you'd rather solve this in a different way. The most robust solution would perhaps be to just copy the entire preamble instead of trying to be clever and extract the list of package names. That would also ensure packages are loaded with the same options as in the document (i.e., `\usepackage[option1,option2]{packagename}`).

### DIFF
--- a/doc/snacks.nvim.txt
+++ b/doc/snacks.nvim.txt
@@ -1,4 +1,4 @@
-*snacks.nvim.txt*            For Neovim           Last change: 2025 October 13
+*snacks.nvim.txt*            For Neovim           Last change: 2025 October 19
 
 ==============================================================================
 Table of Contents                              *snacks.nvim-table-of-contents*
--- a/docs/bigfile.md
+++ b/docs/bigfile.md
@@ -46,6 +46,7 @@ The default implementation enables `syntax` for the buffer and disables
     end
     Snacks.util.wo(0, { foldmethod = "manual", statuscolumn = "", conceallevel = 0 })
     vim.b.minianimate_disable = true
+    vim.b.minihipatterns_disable = true
     vim.schedule(function()
       if vim.api.nvim_buf_is_valid(ctx.buf) then
         vim.bo[ctx.buf].syntax = ctx.ft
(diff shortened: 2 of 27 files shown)


folke:
This is not a proper fix. It's also just not possible to fully fix without actually parsing the doc. Closing

danielwe:
I agree this fix is a poor hack for the problems I described, though it arguably makes the current ad hoc parsing slightly more correct as it avoids the possibility of extracting the wrong argument.

In any case, _some_ ability to tell snacks.image to ignore an `\usepackage` is needed, otherwise the functionality is DOA any time you have a local package. However, fully parsing the preamble doesn't seem worth the complexity and wouldn't really solve this particular problem anyway.

Two alternatives that I think would be more in line with how the package works:
* Adding comment directives a la `% snacks: header start/end`, except they're used for ignoring lines that would otherwise be automatically included. Maybe `% snacks: ignore start/end`
* Dropping the automatic package extraction and telling users to put `% snacks header: start/end` around all parts of the preamble they want to include in the preview, including `\usepackage` lines.

I'd be happy to make a PR for either of these if you're open to it.

PR #2275, opened 2025-10-16
fix(projects): normalize item.text for correct Windows support

dpetka2001:
## Description
`Snacks.picker.projects` would not show correctly projects on Windows. `item.text` was something like `C:\Users\my_user\AppData\Local\nvim-data\lazy\LazyVim\.git\`. 

By normalizing `item.text` it shows correctly the projects.

## Related Issue(s)
None


folke:
ty!

Issue #2271, opened 2025-10-15
feature(terminal): pass count as an argument


drowning-cat:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

1. It is useful for scripting purposes.
2. We cannot use `vim.v.count` directly in terminal mode.

### Describe the solution you'd like

```lua
Snacks.terminal.<open/get/toggle>(nil, { count = 1 })
```

### Related

https://github.com/folke/snacks.nvim/discussions/2268#discussioncomment-14685823

PR #2276, opened 2025-10-16
feat(terminal): minor improvements for user experience

dpetka2001:
## Description
I was trying to create a simple `select` picker to choose terminals to open. 

With the current implementation only the fields `cmd` and `id` are exposed. `cwd` and `env` are not which are used to get the terminal id as well. 

`M.list()` uses `vim.tbl_filter` which turns `terminals` into a list removing the keys terminal ids. Even if that was not the case (i.e using a normal `for` loop to add to `terminals`) the keys would be a string due to `vim.inspect` and one would have to use patterns to extract information from them. 

It's easier to just add `cwd` and `env` to `vim.b[buf].snacks_terminal` to have that information available and be easier for users to manipulate which terminal they want to choose and open.

Also added `opts.count` as an option to be able to programmatically choose the terminal with exactly the same options when it was created, so one can just do 
```lua
local term = vim.b[buf].snacks_terminal
Snacks.terminal(term.cmd, { cwd = term.cwd, env = term.env, count = term.id })
```
to programmatically invoke a certain terminal.

## Related Issue(s)
Solves my issue and #2271 as well.

### DIFF
--- a/lua/snacks/terminal.lua
+++ b/lua/snacks/terminal.lua
@@ -22,6 +22,7 @@ local defaults = {
 
 ---@class snacks.terminal.Opts: snacks.terminal.Config
 ---@field cwd? string
+---@field count? integer
 ---@field env? table<string, string>
 ---@field start_insert? boolean start insert mode when starting the terminal
 ---@field auto_insert? boolean start insert mode when entering the terminal buffer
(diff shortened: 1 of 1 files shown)


folke:
ty!

PR #2281, opened 2025-10-19
feat(toggle): handle table-style options

tkapous:
This PR adds support for creating toggles for table-style options.

The first commit allows toggling the option:

`Snacks.toggle.option('colorcolumn', { on = { '80' }, off = {} })`

The second commit allows toggling sub-options in key-value options:

`Snacks.toggle.option('cursorlineopt', { subopt = 'screenline' })`

Which previously needed to be written like so:

```
Snacks.toggle
  .new({
    id = 'cursorlineopt_screenline',
    name = 'screenline',
    get = function() return vim.tbl_contains(vim.split(vim.o.cursorlineopt, ','), 'screenline') end,
    set = function(state)
      if state then
        vim.opt.cursorlineopt:append('screenline')
        return
      end
      vim.opt.cursorlineopt:remove('screenline')
    end,
  })
```

### DIFF
--- a/lua/snacks/toggle.lua
+++ b/lua/snacks/toggle.lua
@@ -145,24 +145,32 @@ function Toggle:_wk(keys, mode)
 end
 
 ---@param option string
----@param opts? snacks.toggle.Config | {on?: unknown, off?: unknown, global?: boolean}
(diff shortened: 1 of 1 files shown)


folke:

tkapous:
No problem, and thanks for the update!

PR #2288, opened 2025-10-20
fix(dashboard): replace deprecated AutoSession command

franckdelage:
## Description

The command `SessionRestore` from the auto-session plugin is deprecated. Replaced it with `AutoSession restore`.


folke:
ty

PR #2295, opened 2025-10-20
feat(picker): add author field to git log

Aqothy:
## Description

Currently in the picker you cant filter git log pickers by author, the only option to do that right now is to pass it into the opts when you call it like lua Snacks.picker.git_log({ author="test" }) but most of the time I would like to filter interactively and also use the field filtering offered by snacks like file:lua$ and with this new change, author:test.

## Screenshots
Screenshot 2025-10-13 at 11 03 28 PM


folke:
only changed the hl group. ty!

PR #2299, opened 2025-10-21
fix(picker.preview): use `ignorecase` for highlight preview

### LINKED ISSUE #1613: feature(picker): Use ignorecase to highlight matches in preview window (e.g. for grep)
### Did you check the docs?

Hi. I am currently in the process of evaluating snacks.picker as a replacement for telescope. One thing I'd like to enable is ignorecase for preview window of the grep picker (or any comparable picker with a preview pane).

At the moment results in the preview window are only highlighted if they have the same case. However, I'd like all search results to be properly highlighted inside the preview window.

See:

Image

vs:

Image

My expectation would be that the keyword ("Weekly") should be highlighted in both cases

I Don't see any alternative solution

### Additional context

_No response_

dpetka2001:
## Description
As per [comment in 1629](https://github.com/folke/snacks.nvim/pull/1629#issuecomment-2748168988).

## Related Issue(s)
Fixes #1613

### DIFF
--- a/lua/snacks/picker/core/preview.lua
+++ b/lua/snacks/picker/core/preview.lua
@@ -325,7 +325,7 @@ function M:loc()
         hl_group = "SnacksPickerSearch",
       })
     elseif self.filter and vim.trim(self.filter.search) ~= "" then
-      local ok, re = pcall(vim.regex, vim.trim(self.filter.search))
+      local ok, re = pcall(vim.regex, "\\c" .. vim.trim(self.filter.search))
       if ok and re then
         local start = self.item.pos[2]
         local from, to ---@type number?, number?


folke:
## New Approach Implemented

Thanks for working on this! I've implemented a different approach that solves the highlighting issue more comprehensively.

### The Problem
Both PRs were trying to fix highlighting inconsistencies by adding pattern preprocessing (removing grep flags, adding `\c` for case-insensitivity, etc.). However, this approach has limitations:
- **Pattern parsing is fragile** - grep patterns can be complex (regex, fixed-string, case-insensitive combinations)
- **vim.regex quirks** - trying to replicate ripgrep's exact matching behavior in Vim regex is difficult
- **Maintenance burden** - every new grep flag or pattern type needs special handling

### The New Solution
Instead of trying to parse and reinterpret patterns, I've implemented exact match position tracking using ripgrep's built-in features:

**How it works:**
1. Use `rg --replace '__snacks__${0}__snacks__'` to mark exact matches
2. Parse the positions of marked text from ripgrep's output
3. Store positions in `item.positions` array
4. Highlight using exact character indices (no pattern matching needed)

**Benefits:**
- ✅ **100% accurate** - positions come directly from ripgrep, guaranteed to match
- ✅ **Works with any pattern** - regex, fixed-string, case-insensitive, grep flags, etc.
- ✅ **No vim.regex quirks** - no pattern conversion or workarounds needed
- ✅ **Simpler code** - no complex pattern preprocessing logic
- ✅ **Consistent everywhere** - same positions used in list, preview, and file formatter

### Implementation Details
See commit: 3b54c8d

The approach:
- Added `item.positions` field to track match indices
- Modified `grep` source to parse positions from `__snacks__` markers
- Created `highlight.matches()` helper for extmark creation
- Updated list, preview, and format to use positions when available

This supersedes the need for pattern preprocessing since we have exact match positions from the source of truth (ripgrep itself).

---

folke:
Closing in favor of the position-based approach (commit 3b54c8d). Thanks for identifying the issue!

folke:
Let me know if you'd still see issues with this!

dpetka2001:
All seems good. Tnx for quick and better fix 😄

PR #2305, opened 2025-10-21
fix(picker): get rid of side effects when previewing tag locations

dmitmel:
## Description

The current implementation of preview for helptags (and other sources which produce results with tag locations, for that matter) in snacks.picker creates unwanted side effects when browsing the search results, mainly, polluting the search history. This fix changes tag preview code to use the built-in `search()` function for finding the location in the file instead of executing the `/{pattern}` command with `vim.cmd`, which affects neither the search history, nor the jumplist. This is made a special case for tags whose locations begin with `/` or `?`, as outlined in `:help tag-search`. I also add two modifiers to the pattern: `\M` to make sure that it is executed with `'magic'` disabled (again, as described in `:help tag-search`), and `\C` to make the pattern case-insensitive (important so that different locations are previewed for `-l` and `-L`).

### DIFF
--- a/lua/snacks/picker/core/preview.lua
+++ b/lua/snacks/picker/core/preview.lua
@@ -353,8 +353,15 @@ function M:loc()
     end
   elseif self.item.search then
     vim.api.nvim_win_call(self.win.win, function()
-      vim.cmd("keepjumps norm! gg")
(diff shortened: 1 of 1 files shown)


folke:
Your solution didn't highlight the search string.
Found a better way to achieve the same

PR #2322, opened 2025-10-22
feat(picker): add an option to show only the immediate dir name 

SearidangPa:
## Description

  
I have to work in repository with really deep and long directory name. Having an option in the `formatters` for the `dir path` to be only the `immediate dir` would make the UI less noisy for me. I would appreciate any feedback if there is a way to do this better. 

## Screenshots

* Before 
Screenshot 2025-10-21 at 9 49 17 PM

* After
Screenshot 2025-10-21 at 9 50 26 PM


folke:
not interested in adding this. Just set truncate=left

PR #2325, opened 2025-10-23
fix(image): skip `\usepackage` in comments and body

danielwe:
## Description

Here are some small and hopefully uncontroversial tweaks to package extraction from LaTeX preambles:

* Don't consider anything in comments
* Make sure that the extracted names are actually the arguments of `\usepackage` and not some other macro on the same line
* Stop looking for packages at `\begin{document}`, where the preamble ends and the body of the document begins (you can't load packages after this, so any `\usepackage` beyond this point is content, not code. Also saves a huge amount of work in large documents.)


folke:
ty!

PR #2345, opened 2025-10-24
fix(image): avoid nested math environments

danielwe:
## Description

Inline LaTeX equation preview renders equations twice, partly on top of each other, if you use nested math environments. Minimal reproducer:

```latex
\documentclass{article}
\begin{document}
\begin{equation}
  \begin{aligned}
    x_+ &= \frac{-b + \sqrt{b^2 - 4ac}}{2a} \, , \\
    x_- &= \frac{-b - \sqrt{b^2 - 4ac}}{2a} \, .
  \end{aligned}
\end{equation}
\end{document}
```

While, adding a test case for this to `test.tex` I also corrected a package name such that it compiles without error.

## Screenshots

**Before**
Screenshot 2025-10-24 at 16 22 05

**After**
Screenshot 2025-10-24 at 16 23 33

### DIFF
--- a/tests/image/test.aux
+++ b/tests/image/test.aux
@@ -1,6 +1,6 @@
 \relax 
 \@writefile{toc}{\contentsline {section}{\numberline {1}Image Tests}{1}{}\protected@file@percent }
 \@writefile{lof}{\contentsline {figure}{\numberline {1}{\ignorespaces Test image centered in a figure environment.}}{1}{}\protected@file@percent }
-\newlabel{fig:test_image}{{1}{1}{}{}{}}
+\newlabel{fig:test_image}{{1}{1}{}{figure.1}{}}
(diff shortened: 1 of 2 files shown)


AtifChy:
@danielwe how did you enable inline latex equation preview?

danielwe:
Here's the relevant part of my config (I also tweak the latex preamble and the image conversion dpi settings, but that's details):
```lua
  {
    "folke/snacks.nvim",
    optional = true,
    init = function()
      -- inline math preview conflicts with vimtex math mode conceal characters
      vim.g.vimtex_syntax_conceal = {
        greek = 0,
        math_bounds = 0,
        math_delimiters = 0,
        math_fracs = 0,
        math_super_sub = 0,
        math_symbols = 0,
      }
    end,
    opts = { image = { doc = { inline = true, max_width = 1000 } } },
  },
```
Then you need a terminal that supports the kitty graphics protocol. I use Ghostty.

PR #2346, opened 2025-10-25
fix(layout): provide parent win width/height when relative to win

nf-wbeck:
## Description

When a win is relative = "win" it should size itself relative to that win's size. Currently this only happens correctly when the picker layout is created. All subsequent updates, such as a toggle_preview action, can potentially trigger calls to a function width/height opt with a max_width and max_height smaller than was previously calculated. This feeds into a recursive series of updates triggered by the WinResized event triggering and updating with a yet smaller max_width or max_height. 

Additionally, because win.o.cols and win.o.lines is used, relative/decimal dimensions would not calculate correctly in splits. Screenshots are from a picker configured as 
```
{
    col = 1,
    height = 0.9,
    width = 0.9,
    position = 'float',
    relative = 'win',
}
```
**before** 
image

**after**
image

### DIFF
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -251,13 +251,9 @@ function M:update()
     top = (vim.o.showtabline == 2 or (vim.o.showtabline == 1 and #vim.api.nvim_list_tabpages() > 1)) and 1 or 0
   end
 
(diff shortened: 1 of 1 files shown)


folke:
Looks good, but it's not formatted correctly. Please use stylua to format

PR #2354, opened 2025-10-26
feat(image): close hover win on movement if it was opened manually

rodrigoscc:
## Description

For images, I prefer to disable both the `inline` and `float` options. This allows me to use `Snacks.image.hover()` to display the image only when needed. While this works, the hover window does not always close when I move the cursor, it only closes when the cursor moves far from the image. 

I believe most users expect the hover window to close whenever the cursor moves, similar to other hover windows like diagnostics or K. Let me know if you disagree. This PR aims to address this issue and will only affect users who have disabled both the `float` and `inline` options, as they will be using `Snacks.image.hover()` manually to open the hover window.

Thanks for this amazing plugin!

## Screenshots

Before:

before

After:

after

### DIFF
--- a/lua/snacks/image/doc.lua
+++ b/lua/snacks/image/doc.lua
@@ -370,6 +370,8 @@ function M.hover()
   local current_win = vim.api.nvim_get_current_win()
   local current_buf = vim.api.nvim_get_current_buf()
 
+  local float = Snacks.image.config.doc.float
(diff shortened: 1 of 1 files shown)


folke:
If you manually trigger hover, then you also need to close it manually.
There will be users that don't want to auto-close

PR #2366, opened 2025-10-28
fix(picker.git_diff): properly parse oneline and combined hunks

### LINKED ISSUE #1834: bug: git picker source fails `diff` when Git `diff.mnemonicprefix` set

### Neovim version (nvim -v)

0.11.1

### Operating system/version

MacOS 15.4.1

### Describe the bug

This line tries to match on `a/` and `b/` prefixes, however when `diff.mnemonicprefix` ([git-scm.com/diff-config#diffmnemonicPrefix](https://git-scm.com/docs/diff-config#Documentation/diff-config.txt-diffmnemonicPrefix)) is set, different prefixes are used (`i/`, `w/`, `c/`, or `o/` depending on context).
https://github.com/folke/snacks.nvim/blob/bc0630e43be5699bb94dadc302c0d21615421d93/lua/snacks/picker/source/git.lua#L241

This causes `file` to be `nil` and the `unexpected line` error to be thrown with the index line

### Steps To Reproduce

```
git config --global diff.mnemonicprefix true
```

### Expected Behavior

No error should be thrown, regardless of `diff.mnemonicprefix` config value.

shizukayuki:
## Description

Fix the parsing of merge conflict and 1 line file diffs, alongside some changes to make it more robust.

## Related Issue(s)

* Fixes #1834

### DIFF
--- a/lua/snacks/picker/source/git.lua
+++ b/lua/snacks/picker/source/git.lua
@@ -256,14 +256,13 @@ end
 ---@param opts snacks.picker.git.diff.Config
 ---@type snacks.picker.finder
 function M.diff(opts, ctx)
-  local args =
-    M.git("diff", "--no-color", "--no-ext-diff", { args = { "-c", "diff.noprefix=false", "--no-pager" } }, opts)
+  local args = M.git("diff", "--no-color", "--no-ext-diff", "--default-prefix", { args = { "--no-pager" } }, opts)
   if opts.base then
     vim.list_extend(args, { "--merge-base", opts.base })
   end
   local file, line ---@type string?, number?
   local header, hunk = {}, {} ---@type string[], string[]
-  local header_len = 4
+  local header_len = 0
 
   local cwd = svim.fs.normalize(opts and opts.cwd or uv.cwd() or ".") or nil
   cwd = Snacks.git.get_root(cwd) or cwd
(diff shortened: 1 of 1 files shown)


folke:
I was actually in the process of refactoring the diff code, so I couldn't merge this.
I did however take some of the logic in the new parser. Thanks!

shizukayuki:
had it locally for awhile now, so nice to have it upstream.

3049ad8beba417924b2d4ff8f77edec53d3472f6 seems fine, other than it failing on filenames with spaces
since `util.parse_cmdline` can't parse this or a double space correctly.
```diff
diff --git a/READ ME b/READ ME
--- a/READ ME	
+++ b/READ ME	
```

i think this is why i used the path from `--- a/` as it will only fail if a filename has a `\t` or `\n`
```diff
diff --git "a/word1\tword2" "b/word1\tword2"
--- "a/word1\tword2"
+++ "b/word1\tword2"
```

folke:
Is this valid though `diff --git a/READ ME b/READ ME`?

It should have `""` in this case?

shizukayuki:
both are valid, it is only quoted if a char needs to be escaped.

shizukayuki:
output of `git diff`
```diff
diff --git i/file 1 w/file 1
index 3367afd..3e75765 100644
--- i/file 1	
+++ w/file 1	
@@ -1 +1 @@
-old
+new
```

and `diff -u file\ 2 file\ 1`
```diff
--- "file 2"	2025-10-28 13:06:33.225359975 +0400
+++ "file 1"	2025-10-28 13:06:07.057488731 +0400
@@ -1 +1 @@
-old
+new
```

folke:
Made some changes. Let me know if it's still not ok

shizukayuki:
the only edge case i can think of is `"dir A/file"` which outputs `diff --git i/dir A/file w/dir A/file`

also can revert the `--default-prefix` change since parsing is no longer reliant on prefix after refactor.
edit: nvm `git diff --dst-prefix=dst README` prints `+++ dstREADME`

PR #2368, opened 2025-10-28
fix(picker.actions): avoid reloading buffer if it's the current buffer

Davidyz:
## Description

When using `tabdrop` as the `jump` action and jumping to a location in the current buffer, the existing implementation always triggers a reload of the current buffer because the `vim.cmd` is unconditional. This causes unnecessary triggering of `BufRead*` and `LspAttach` auto-commands, which can then trigger a lot of things that can be CPU-intensive (linters, LSP, etc.). This PR skips calling the `cmd` when jumping to the current buffer, and hence avoids the excessive triggering of those auto-commands.

### DIFF
--- a/lua/snacks/picker/actions.lua
+++ b/lua/snacks/picker/actions.lua
@@ -53,7 +53,7 @@ function M.jump(picker, _, action)
   local win = vim.api.nvim_get_current_win()
 
   local current_buf = vim.api.nvim_get_current_buf()
-  local current_fname = nil
+  local current_tab = vim.api.nvim_get_current_tabpage()
   local current_empty = vim.bo[current_buf].buftype == ""
     and vim.bo[current_buf].filetype == ""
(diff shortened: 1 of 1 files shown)


folke:
Just don't use tabdrop in this case?

Davidyz:
> Just don't use tabdrop in this case?

Tabdrop works fine when jumping to other buffers. Am I supposed to create a special binding _just_ for jumping to somewhere in the current buffer? Sounds very counterintuitive to me (and not ergonomic either).

Davidyz:
FYI @folke, the current implementation (without my patch) also throws an error if there are unsaved changes in the current buffer and tabdrop goes to the current buffer/tab.

folke:
I've simplified the jump function to no longer use the actual `:drop` command.
It now does the same as drop, but will never reload existing buffers.

PR #2377, opened 2025-10-29
feat(picker.explorer): allow sorting when search query is empty

### LINKED ISSUE #2376: feature: allow to access explorer.item.sort when matcher.sort_empty is true
### Did you check the docs?

I have this to achieve [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order) for exlorer picker:
```lua
        explorer = {
          --- @type snacks.picker.sort
          sort = function(a, b)
            local av = a.sort
            local bv = b.sort

            if av ~= nil and bv ~= nil then
              -- http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua
              local function padnum(d)
                local dec, n = string.match(d, "(%.?)0*(.+)")
                return #dec > 0 and ("%.12f"):format(d) or ("%s%03d%s"):format(dec, #n, n)
              end
              return tostring(av):gsub("%.?%d+", padnum) .. ("%3d"):format(#bv) <
                  tostring(bv):gsub("%.?%d+", padnum) .. ("%3d"):format(#av)
            end
            return false
          end,

          matcher = { sort_empty = true },
        },
```

It requires `item.sort`, it will get generated on `explorer.search(opts, ctx)` function.

But 

https://github.com/folke/snacks.nvim/blob/59c5545e91878c1f6218b032a881832bc98a46f3/lua/snacks/picker/source/explorer.lua#L196-L199

will return `explorer.search(opts, ctx)` only when `state:setup(ctx)` is true, it means search queue should have something.
Can you make explorer returns M.search if `matcher = { sort_empty = true }`?

Modify the section like this
```lua
  if state:setup(ctx) or opts.matcher.sort_empty then
    ctx.picker.matcher.opts.keep_parents = true
    return M.search(opts, ctx)
  end
```

idk

### Additional context

I will make pull request soon

zknx:
## Description

Allow sorting when search query is empty by using `explorer.search(opts, ctx)` when `opts.matcher.sort_empty` is true, currently I can't sort files if query is empty (sorting by like `item.path` makes tree structure broken)

## Related Issue(s)

- Fixes #2376

### DIFF
--- a/lua/snacks/picker/source/explorer.lua
+++ b/lua/snacks/picker/source/explorer.lua
@@ -193,7 +193,7 @@ function M.explorer(opts, ctx)
   local state = M.get_state(ctx.picker)
 
   ctx.picker.matcher.opts.keep_parents = false
-  if state:setup(ctx) then
+  if state:setup(ctx) or opts.matcher.sort_empty then
     ctx.picker.matcher.opts.keep_parents = true
     return M.search(opts, ctx)
   end


folke:
Yeah no, that's for sure not correct. That's the switch between search and explore mode.

PR #2402, opened 2025-11-01
fix(picker.git_diff): parsing of filenames and renames

shizukayuki:
## Description

followup to #2366 because current code can't parse `diff --git i/dir a/file w/dir a/file` correctly.
* bring back `diff.noprefix=false` since parsing diff cmd without prefixes is not reliable
* fix renames using old name instead of new name
* add `--diff-filter=u` to filter out `* Unmerged path ...` from output

### DIFF
--- a/lua/snacks/picker/source/diff.lua
+++ b/lua/snacks/picker/source/diff.lua
@@ -91,6 +91,9 @@ function M.parse(lines)
       hunk = nil
     end
     if block then
+      if block.file:sub(1, 1) == '"' then
+        block.file = block.file:sub(2):sub(1, -2)
+      end
       table.sort(block.hunks, function(a, b)
         return a.line < b.line
       end)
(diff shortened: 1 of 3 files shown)


folke:
I just pushed a bunch of changes to the diff parsing. Should be way more robust now.
The additional tests in this PR were very helpful!

Blocks now also have a type (delete, rename, etc) and properly parse the left/right filenames.

`noprefix` diff output should also be handled correctly.



Will also update the diff picker to show this extra information correctly.

folke:
Relevant changes here https://github.com/folke/snacks.nvim/commit/b6e4eb7e608924a0d116bb01dabb575365bcfd75

shizukayuki:
looking at b6e4eb7e608924a0d116bb01dabb575365bcfd75 only things that seems off are:
* `block.type` is dependent on the header order. a block can be both `mode` and `rename`
  ```diff
  diff --git c/bin/script.sh i/bin/legacy/script.sh
  old mode 100755
  new mode 100644
  similarity index 100%
  rename from bin/script.sh
  rename to bin/legacy/script.sh
  ```
* rename/copy headers shouldn't call `norm` because prefixes are not used

also much prefer setting `noprefix=false` over hardcoded prefix list, not so important since the func mostly deals with git diffs and having a root directory match a prefix is probably rare.

folke:
All fixed as well!

Next up is new diff rendering similar to what I did for gh pr reviews and similar to delta.

Once I get that down, I'll start integrating gh reviews in GitHub PR Diffs + making it easy to do reviews on diffs.

shizukayuki:
thank you for your work!

not a user of either, but hope that turns out well.

folke:
Sneak preview

image

shizukayuki:
it does look nice, but i like my diffs with less colors anyway.

and i'm half expecting this to fail on my setup since my usual colorscheme is catppuccin.
maybe should file a bug against catppuccin when it is done.

PR #2435, opened 2025-11-04
feat(gh): add gh enterprise support

thomasbecker:
## Description

Add support for github enterprise cloud and on premise installations.

## Related Issue(s)

n/a

## Screenshots

n/a

### DIFF
--- a/doc/snacks.nvim-image.txt
+++ b/doc/snacks.nvim-image.txt
@@ -165,7 +165,7 @@ In case of issues, make sure to run `:checkhealth snacks`.
       },
       ---@class snacks.image.convert.Config
       convert = {
-        notify = true, -- show a notification on error
+        notify = false, -- show a notification on error
         ---@type snacks.image.args
         mermaid = function()
           local theme = vim.o.background == "light" and "neutral" or "dark"
--- a/doc/snacks.nvim-picker.txt
+++ b/doc/snacks.nvim-picker.txt
@@ -854,7 +854,7 @@ TROUBLE                                  *snacks.nvim-picker-examples-trouble*
 >lua
     ---@alias snacks.picker.format.resolve fun(max_width:number):snacks.picker.Highlight[]
     ---@alias snacks.picker.Extmark vim.api.keyset.set_extmark|{col:number, row?:number, field?:string}
-    ---@alias snacks.picker.Text {[1]:string, [2]:(string|string[])?, virtual?:boolean, field?:string, resolve?:snacks.picker.format.resolve}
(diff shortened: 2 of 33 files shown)


folke:
Can you tell me what failed without this PR?
There's other users with GHE that can use it without this PR.

thomasbecker:
> Can you tell me what failed without this PR? There's other users with GHE that can use it without this PR.

Ah interesting. You switched to M.get_repo() which makes it work if GH_HOST is set. That didn't work when I started on this PR as github.com was hardcoded.

My variant works always works without setting GH_HOST which is quite handy when working with different ghe instances. 

Should I have a closer look at your latest changes and we merge them together?

thomasbecker:
Just had a closer look. As said main works now if GH_HOST is set as we basically just call gh and it reads the host from the env variable.

My variant does that, too. But even without GH_HOST being set it should always be able to autodetect and set it. I've multiple projects with different ghe instances. That way I don't have to care for anything, it just always works. WDYT?

folke:
The easier approach seems to be to just `:h 'exrc'` for this?

Create a file in the project `.nvim.lua` and just set `GH_HOST` there?

thomasbecker:
I appreciate the concern about keeping things simple, but I'd like to make a case for why auto-detection provides value over the .nvim.lua approach:

The .nvim.lua approach has friction:

1. Requires creating a file in every repository that uses GHE
1. Can't be committed to the repo (security - exposes internal host names)
1. Needs manual setup for every clone/machine
1. Breaks the "just works" experience when switching between repos with different GHE instances

Auto-detection eliminates all of this:

- Zero configuration needed
- Works immediately on every clone
- Automatically adapts when switching between github.com and GHE repos
- The git remote URL already contains the host information - we're just reading what's already there

The implementation is straightforward:

- ~80 lines in a single host.lua module
- Uses standard Lua io.popen() to read git config (callback-safe)
- Caches the result to avoid repeated git calls
- Priority chain: explicit config > auto-detect > GH_HOST env var > github.com

Compromise proposal:
Make it opt-in via config if you prefer:
gh = {
auto_detect_host = true, -- defaults to false if you prefer
}

This way users who want the convenience can enable it, and it doesn't affect anyone who doesn't need it.

For users working with multiple GHE instances (which is increasingly common in enterprises with acquisitions/separate divisions), auto-detection turns a recurring manual task into something that "just works."

I love KISS, but this would make working with many different GHE instances really more convenient at the cost of little complexity.

WDYT?

folke:
If anything, this should be fixed in `gh` cli.
I don't see why I should add a work-around for this in snacks.

thomasbecker:
Reasonable and finally your decision. :)

As always thx for all the work you put into the neovim ecosystem.

Issue #1234, opened 2025-02-17
`:h snacks` should go to highest level help page


kaddkaka:
### Did you check the docs?

- [x] I have read all the snacks.nvim docs

### Is your feature request related to a problem? Please describe.

Currently `:h snacks` goes to `snacks-dim.txt`,  a very specific help page.

### Describe the solution you'd like

Arrive at  `snacks.nvim.txt`

### Describe alternatives you've considered

Is there a more general top level page that I should be brought to instead?

PR #2443, opened 2025-11-05
fix(gh_actions): missig pr context when picker launched standalone for web and yank actions

### LINKED ISSUE #2442: bug: gh web browser action has missing PR

### Neovim version (nvim -v)

0.12.0-dev-1565+gb80d390765

### Operating system/version

Mac OS 15.7

### Describe the bug

When invoking `Snacks.picker.gh_actions()` directly (without first selecting a PR from `Snacks.picker.gh_pr()`), choosing the "Open in web browser" action shows a notification: `Opened #nil in web browser`. The correct PR number is displayed in the action list itself, so the underlying PR was detected; only the notification loses the number in this invocation path. Other actions (e.g. "View PR diff") work correctly.

### Steps To Reproduce

1. Open Neovim inside a git repository with an active branch that has an associated PR on GitHub.
2. Run: `:lua Snacks.picker.gh_actions()`
3. In the actions picker that appears, select "Open in web browser".
4. Observe the notification: `Opened #nil in web browser`.

### Expected Behavior

Notification should display `Opened #<PR number> in web browser` (e.g. `Opened #123 in web browser`).

### Repro

```lua
require("lazy.minit").repro({
  spec = {
  },
})
-- After startup: :lua Snacks.picker.gh_actions()
```

aweis89:
## Description
Fix `gh_browse` action producing `Opened #nil in web browser` when invoked from the standalone `Snacks.picker.gh_actions()` picker. A similar issue happened with the yank action as well.

## Root Cause
In the standalone actions picker, entries are wrapper tables (`{ action = ..., item = <gh item> }`). The `gh_browse` action iterated `ctx.items` assuming raw GitHub items (`it.number`), so `number` was `nil` and the notification showed `#nil` even though the browser opened the correct URL. The `gh_yank` action likewise assumed unwrapped items and attempted to read `it.url` directly, resulting in empty or missing URLs for wrapped entries.

## Changes
- Add `unwrap` helper and use it.
- Detect and unwrap items in `gh_browse`.
- Support acting on all selected items (falls back to single item).
- Clear list selection after action to avoid accidental repeats.
- Update `gh_yank` to unwrap items, ignore nils, and warn when no URLs.
- Preserve existing behavior for list pickers (no regressions).

## Behavior Before
Running `:lua Snacks.picker.gh_actions()` then choosing "Open in web browser" on a PR/issue showed `Opened #nil in web browser`. Yanking URLs from the standalone actions picker could produce no URLs or incorrect results.

## Behavior After
- Correct notification: `Opened #1234 in web browser`.
- Multi-select from list picker opens each selected item’s URL with individual notifications.
- Yanking from wrapped selections copies the correct URL list (or warns if none).
- Selection cleared after actions.

## Issue
Closes #2442

### DIFF
--- a/lua/snacks/gh/actions.lua
+++ b/lua/snacks/gh/actions.lua
@@ -123,15 +123,18 @@ M.actions.gh_perform_action = {
   end,
 }
 
+local function unwrap(it)
+  return it and (it.number and it or it.item) or nil
+end
+
 M.actions.gh_browse = {
   desc = "Open in web browser",
   title = "Open {type} #{number} in web browser",
   icon = " ",
   action = function(item, ctx)
-    -- Normalize items: in list pickers ctx.items are GH items; in actions picker they are wrappers.
     local items = (ctx.items and #ctx.items > 0) and ctx.items or { item }
     for _, it in ipairs(items) do
-      local gh = it.number and it or it.item -- unwrap if needed
+      local gh = unwrap(it)
       if gh and gh.number then
         Api.cmd(function()
           Snacks.notify.info(("Opened #%s in web browser"):format(gh.number))
@@ -142,7 +145,7 @@ M.actions.gh_browse = {
       end
     end
     if ctx.picker then
-      ctx.picker.list:set_selected() -- clear selection
+      ctx.picker.list:set_selected()
     end
   end,
 }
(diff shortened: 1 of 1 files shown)


folke:
Thank you for letting me know. The real fix was to update `gh_perform_action` and pass the correct context, so that it works for all actions.

PR #2459, opened 2025-11-06
fix(picker.lsp_config): nil on lsp info

EasonMo:
## Description

It will report message `error: vim/shared.lua:0: t: expected table, got nil` when execute `Snacks.picker.lsp_config()`.


folke:
What Neovim version are you on where that doesn't exist?

dpetka2001:
@folke It exists only on nightly. On 0.11.4 and 0.11.5 it doesn't exist. The equivalent there is `_request_name_to_capability`.

PS: reference from Neovim https://github.com/neovim/neovim/blob/b92e92b20dfcebfaa32d2f3265449d1306a66d4b/runtime/lua/vim/lsp/protocol.lua#L1089

folke:
ok, good to know., thanks!

PR #2501, opened 2025-11-11
feat(layout): add layout level footer keys

### LINKED ISSUE #2469: feature: footer keys on  the picker
### Did you check the docs?

Snacks.picker is now a central piece of my Neovim setup.
However, some pickers that I use less often have some extra keybindings I struggle to remember.

It would be great to have a `footer_keys` option enabling a footer with the keys, but something a bit more dynamic than the `win.footer_keys` that would display the keys for the current active window (input, list, preview) below the current layout.

Image

Already tried using:
- `picker.win.input.footer_keys`: break the display (to many keys to display for the size)
- `picker.win.list.footer_keys`: nothing displayed
- `picker.win.preview.footer_keys`: currently enabled in my config, only display the preview shortcut

This approach makes many keys in a small place (in the case of half listing, half preview), displays duplicate common keys (cancel, cycle_win...), and is not contextual (displayed for each part of the layout, focused or not).

### Additional context

_No response_

noirbizarre:
## Description

I tried to implement #2469 with the idea I had in mind. So here's my proposal

This PR adds support for layout level footer keys display (keys for the current active window). If the root box has no border, a transparent bottom border is dynamically added to display the footer keys.

It also adds to both window and layout footer key filtering on the current mode.

The `select` picker is excluded to keep it simple.

It initially started with only layout level footer keys, then I've been down the rabbit hole to provide a polished solution (IMHO):

- Adds a `max_footer_keys` option at both `win` and `layout` to cap the number of displayed keys (avoid keys being cropped)
- Adds an optional `weight` field on keys so the most important (the bigger weight) are displayed first. Added a weight to default existing keys (probably needs review; helped with AI for mass weight addition).
- Fix action description not being displayed when available. Added description to all keys in the codebase pointing to actions without description.

I would love a feedback this, if you are interested.

## Related Issue(s)

- Fixes #2469

## Screenshots

Result in the picker with:

```lua
picker = {
    layout = {
        footer_keys = true,
        footer_max_keys = 10,
    },
}
```

Capture d’écran du 2025-11-11 13-30-23

### DIFF
--- a/lua/snacks/layout.lua
+++ b/lua/snacks/layout.lua
@@ -609,20 +609,8 @@ function M:update_footer()
   if not focused then
     return
   end
-  -- Determine effective max keys (minimum of layout and window caps if both set)
-  self.root.opts.footer = focused:build_footer_keys(want, self.opts.footer_max_keys)
-  if self.root:valid() then
-    local cfg = vim.api.nvim_win_get_config(self.root.win)
-    cfg.footer = self.root.opts.footer
-    cfg.footer_pos = self.root.opts.footer and (self.root.opts.footer_pos or "center") or nil
-    -- Ensure root has a bottom border so we can show a footer
-    -- Only auto-add a bottom border if user did not explicitly set any border
-    if not self.root:has_border() and not self._root_border_explicit then
-      cfg.border = { "", "", "", "", "", " ", "", "" }
-    end
-
-    pcall(vim.api.nvim_win_set_config, self.root.win, cfg)
-  end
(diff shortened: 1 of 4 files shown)


noirbizarre:
I've been using it the whole week, and it works really great. I even (re)discovered some features I wasn't aware of or I just forgot, so (to me) it totally serves its discovery purpose.
However, I might add an extra commit to adjust the weights. This is not a blocker, just refinement. And I think, it would need some external advice/feedback too, as there is a part of personal taste in the weight choice.
So anyone seeing this PR and being interested, feedback would be greatly appreciated. 🙏🏼

PR #2517, opened 2025-11-15
feat(picker.highlight): use SnacksPickerTitle hl group

willnorris:
Allow styling the title of the picker window directly. This still defaults to `FloatTitle`, so this change has no effect if `SnacksPickerToggle` hasn't been directly set.

Without this, the only way to adjust the Picker window title is to set `FloatTitle` which affects all floating windows.a

### DIFF
--- a/lua/snacks/picker/core/picker.lua
+++ b/lua/snacks/picker/core/picker.lua
@@ -404,7 +404,7 @@ function M:update_titles()
           return a:sub(1, 1):upper() .. a:sub(2)
         end, vim.split(name, "_")))
         toggles[#toggles + 1] = { " " .. toggle.icon .. " ", "SnacksPickerToggle" .. hl }
-        toggles[#toggles + 1] = { " ", "FloatTitle" }
+        toggles[#toggles + 1] = { " ", "SnacksPickerTitle" }
       end
     end
   end
@@ -415,7 +415,7 @@ function M:update_titles()
     if win.opts.title then
       local tpl = win.meta.title_tpl or win.opts.title
       win.meta.title_tpl = tpl
-      tpl = type(tpl) == "string" and { { tpl, "FloatTitle" } } or tpl
+      tpl = type(tpl) == "string" and { { tpl, "SnacksPickerTitle" } } or tpl
       ---@cast tpl snacks.picker.Text[]
 
       local has_flags = false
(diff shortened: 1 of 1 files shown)


anthraxx:
I stumbled upon his issue while trying to debug the same issue: Can't style the picker title fully individually (per component).

The issue with this PR though, is that this seems to conflict the way `SnacksPickerPreviewTitle` is respected, with the PR applied I can no longer set the preview title highlight to something else than the actual picker input title.

anthraxx:
@willnorris Okay I figured the following, which may work for you:
with the regular `default` layout, you can use `SnacksTitle` but *NOT* `SnacksPickerTitle` and override the preview one with `SnacksPickerPreviewTitle`
when changing layout to `telescope` you can use the more specific `SnacksPickerTitle` alongside `SnacksPickerPreviewTitle` and `SnacksPickerListTitle`

Taking this into account, the question is probably like why `SnacksPickerTitle` has no effect with the `default` layout.

EDIT: I assume this is all actually expected behavior. The `default` layout uses the `title` property on the root of the picker layout, while the `telescope` one specifies the `title` property on the `input` win. That said, I'm still a bit curious why  `SnacksPickerTitle` doesn't apply to the `title` on the root of the layout but `SnacksTitle` does.

willnorris:
okay,  you're definitely right that this was the wrong approach.  I'm actually using the "ivy" layout, and I think I've discovered that this problem exists with any layout that sets a title at the root window rather than on an inner window.  Of the built-in layouts, that includes ivy, ivy_split, and vertical.  For example, ivy is:

```lua
M.ivy = {
  layout = {
    box = "vertical",
    backdrop = false,
    row = -1,
    width = 0,
    height = 0.4,
    border = "top",
    title = " {title} {live} {flags}",
    title_pos = "left",
    { win = "input", height = 1, border = "bottom" },
    {
      box = "horizontal",
      { win = "list", border = "none" },
      { win = "preview", title = "{preview}", width = 0.6, border = "left" },
    },
  },
}
```

In this case, the title on the root window is not using the `SnacksPickerTitle` highlight group.

If instead you move the title into the input window, and adjust the borders accordingly (like most of the other layouts do):

```lua
M.ivy = {
  layout = {
    box = "vertical",
    backdrop = false,
    row = -1,
    width = 0,
    height = 0.4,
    border = "none",
    { win = "input", height = 1, border = "top_bottom", title = " {title} {live} {flags}", title_pos = "left" },
    {
      box = "horizontal",
      { win = "list", border = "none" },
      { win = "preview", title = "{preview}", width = 0.6, border = "left" },
    },
  },
}
```

Then it properly applies the SnacksPickerInputTitle highlight group.  That's what I've done for now, and it works well.

I'm sure the root issue is how the highlight groups are getting applied to these windows, but I'm having trouble tracking down where the issue is.

PR #2622, opened 2025-12-14
feat(picker.actions): configurable chars for pick_win

Blackmorse:
## Description

Previously, the `pick_win` action in `snacks.nvim` used a hardcoded string (`"asdfghjkl"`) to label picking windows. This PR adds the ability to customize the label string, allowing users to specify their preferred characters (e.g., `"fjdksla;cmrueiwoqp"`, as it is by default in [nvim-window-picker](github.com/s1n7ax/nvim-window-picker) plugin).


metziger23:
I would like to have this as well

PR #2623, opened 2025-12-14
feat(picker actions): picker_grep accepts picker to search selected dirs

joshzcold:
## Description

- this allows selecting directories in `exporter` and doing `<leader>/`
  to search the selected dirs from the user. This still supports a
  single item from the explorer as well.
  
  This is helpful in mono-repo setups where usually to get "scoped search" you would need to `:cd <dir>` then search which removes the context from the sibling repos. Being able to `<leader>/` in `Snacks.explorer` is pretty close, but I think this feature is one step further in nice functionality.

## Related Issue(s)

> None that I can find. 

## Screenshots

1. Selecting a few directories to search 

image

2. Pressing `<leader>/` to show results just from those 2 selected directories

image

## Questions

1. Does it make sense to deselect from these directories in explorer after the search is done or is it more useful to keep them selected for next searches

2. Is there a way to pass the `selected` to a separated search outside of `explorer` like `Snacks.picker.files()`

3. Is this concept worth doing the ground work to have different types of `selected` that don't conflict with `copy/paste` ? Other IDEs let you enable and disable "folders" within a project to scope search. Could be next to have that also be a supported feature in Snacks.picker.


joshzcold:
To answer question number 2, this works:

```lua
      { "<leader>e",      function()
        local pickers = Snacks.picker.get({source= "explorer"})
        local picker = pickers[1]

        local search_dirs = {}
        for _, selected in ipairs(picker:selected({ fallback = true })) do
          table.insert(search_dirs, Snacks.picker.util.path(selected))
        end

        if search_dirs then
          Snacks.picker.files({ dirs = search_dirs })
        else
          Snacks.picker.files()
        end

      end, desc = "Find Config File" },
```

This works for getting the active picker.

Considering that this works while the explorer is open, it might not be needed to overwrite the `<leader>/` behavior of `picker_grep` considering that I could achieve the same result in my config?

Maybe this is better stored in a recipe?

Thoughts?

Issue #2634, opened 2025-12-21
bug: images don't re-render when re-entering a buffer


afawe3331:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.12.0-dev

### Operating system/version

macOS 26.3

## Description

Inline images in markdown files are not re-rendered when re-entering a buffer after switching to another buffer. The images only appear again after manually reloading the file (`:edit` or closing and reopening).

## Steps To Reproduce

1. Open a markdown file with inline images using snacks.nvim with `doc.inline = true`
2. Images render correctly initially
3. Switch to another buffer (`:bn`, `:e otherfile.md`, etc.)
4. Switch back to the original markdown buffer (`:bp`)
5. **Expected**: Images should appear immediately
6. **Actual**: Images do not appear until the file is reloaded

## Environment

- Neovim version: 0.12.0-dev
- OS: macOS 26.3
- Terminal: Kitty
- snacks.nvim commit: fe7cfe9800a182274d0f868a74b7263b8c0c020b

### Checkhealth Output

```
Snacks.image ~
- ✅ OK setup {enabled}
- ✅ OK 'kitty' `kitty 0.44.0 created by Kovid Goyal`
- ✅ OK 'magick' `Version: ImageMagick 7.1.2-11 Q16-HDRI aarch64 23470 https://imagemagick.org`
- ✅ OK 'convert' `WARNING: The convert command is deprecated in IMv7, use "magick" instead of "convert" or "magick convert"`
- ✅ OK `kitty` detected and supported
- ✅ OK `kitty` supports unicode placeholders
- ✅ OK Inline images are available
- ✅ OK Terminal Dimensions:
  - {size}: `1710` x `2136` pixels
  - {scale}: `1.88`
  - {cell}: `15` x `24` pixels
- ✅ OK Available Treesitter languages:
    `html`, `javascript`, `markdown_inline`, `markdown`, `tsx`
- ⚠️ WARNING Missing Treesitter languages:
    `css`, `latex`, `norg`, `scss`, `svelte`, `typst`, `vue`
- ⚠️ WARNING Image rendering in docs with missing treesitter parsers won't work
- ❌ ERROR Tool not found: 'gs'
[...]

PR #2635, opened 2025-12-21
fix(image): restore image visibility when buffer returns to view

afawe3331:
When a buffer with an image was hidden (no open windows), and then shown again, the image would not render
   because the 'hidden' flag was never restored to false.

     This fix calls show() when the buffer becomes visible again, properly restoring the image visibility state.

     ## Description

     When switching between buffers in Neovim with snacks.nvim's image viewer enabled:
     1. Open an image file - image displays correctly ✅
     2. Switch to another buffer - the image buffer has no visible windows
        - `M:update()` calls `self:hide()` and sets `self.hidden = true`
     3. Return to the image buffer - the buffer is now visible again
        - `M:update()` is called with `#state.wins > 0` ✅
        - But `self.hidden` remains `true` ❌
     4. The `_render()` function checks `if self.hidden then` and renders empty text instead of the image ❌

     **Root Cause**: The `hidden` flag is set when the buffer is hidden, but never restored when it becomes
   visible again.

     **Fix**: In `M:update()`, after checking that windows exist (`#state.wins > 0`), check if the buffer was
   previously hidden and call `show()` to restore the visibility state.

     ## Related Issue(s)

     This is a fix for image disappearance when switching between buffers in snacks.nvim.

     ## Changes Made

     - Modified `lua/snacks/image/placement.lua`
     - Added 3 lines in `M:update()` to restore `show()` when buffer becomes visible

     ## Testing

     ✅ Tested on macOS with Kitty terminal
     - Open PNG/JPG image file → image displays
     - Switch to another buffer → buffer hidden correctly
     - Return to image buffer → image now displays correctly (previously was blank)
     - Repeated multiple times → consistently working

     ## Code Change

     ```lua
     if #state.wins == 0 then
       self:hide()
       return
     end
     if self.hidden then
       self:show()  -- ← Added: restore visibility when buffer becomes visible
     end
     self.img:place(self)

### DIFF
--- a/lua/snacks/image/placement.lua
+++ b/lua/snacks/image/placement.lua
@@ -538,6 +538,9 @@ function M:update()
     self:hide()
     return
   end
+  if self.hidden then
+    self:show()
+  end
   self.img:place(self)
 
   self:debug("update")


afawe3331:
Fixes #2634

yurijmikhalevich (approved):

yurijmikhalevich:
@folke, hi! I can reproduce the issue with the latest snacks.nvim. Do you think we can get this merged? Thank you!

yurijmikhalevich:
If someone is finding this thread dealing with the same issue, adding this init function to the `snacks.nvim` config is a workaround:

```lua
return {
  "folke/snacks.nvim",
  opts = {
    image = {
      enabled = true,
      doc = {
        enabled = true,
      },
      math = {
        enabled = true,
      },
    },
  },
  init = function()
    -- Workaround: Force Snacks to re-render images on buffer switch
    vim.api.nvim_create_autocmd("BufWinEnter", {
      pattern = { "*.png", "*.jpg", "*.jpeg", "*.gif", "*.webp", "*.avif" },
      callback = function(args)
        vim.schedule(function()
          -- Ensure the buffer is still valid before applying the setting
          if vim.api.nvim_buf_is_valid(args.buf) then
            -- Unload the buffer when hidden. When you switch back, 
            -- Neovim is forced to reload it, re-triggering the Snacks viewer.
            vim.bo[args.buf].bufhidden = "unload"
          end
        end)
      end,
    })
  end,
}
```

PR #2647, opened 2025-12-29
fix(latex): latex inline + hover

junyixu:
## Screenshots

image


junyixu:
Previously,  after a LuaSnip expansion(entering Select mode), the inline preview image ends up covering the code(i.e. latex) content.

image

The luasnip:
```lua
parse({ trig = "sum", name = "sum" }, "\\sum_{n=${1:1}}^{${2:\\infty}} ${3:a_n z^n}"),
```

PR #2685, opened 2026-01-22
fix(lazygit): nvim opening over itself

BirdeeHub:
Makes workarounds like this no longer necessary:

https://github.com/BirdeeHub/nix-wrapper-modules/blob/48b4a74b16a800f57c4ef4c1b10cfe46bb7847d3/templates/neovim/init.lua#L306-L347

I have had this fixed for a few months in my own config with this method and it has solved my annoyance. Realized it was easier to extract than I thought.


BirdeeHub:
The defaults for these may also currently just be wrong in lazygit.

```bash
[ -z "$NVIM" ] && (nvim -- {{dir}}) || (nvim --server "$NVIM" --remote-send "q" && nvim --server "$NVIM" --remote-tab {{dir}})
```

I think they have their logic wrong here, the () don't look like they are doing what the author intended.

https://github.com/jesseduffield/lazygit/blob/19605ad47699888da471ddf17bd2ae6dbb1abf3a/pkg/config/editor_presets.go#L64-L66

Regardless, using `vim.v.progpath` has guarantees that just using the nvim command does not, but they do not have access to `vim.v.progpath` inside lazygit.

As such we should still implement these here, regardless of if their defaults for those options are currently bugged or not.

I did not set [editAtLineAndWait](https://github.com/jesseduffield/lazygit/blob/19605ad47699888da471ddf17bd2ae6dbb1abf3a/pkg/config/editor_presets.go#L76) because it is not currently a remote command and technically is not yet supported, it is fine as just nvim for that one for now.

But when that one is supported and becomes a remote command, I would expect that to be done here as well.

PR #2687, opened 2026-01-23
fix(picker): correct Windows platform check for the `find` command

glmlm:
## Description

Fix invalid enable check for the `find` command that always evaluated to true.

## Related Issue(s)

N/A

## Screenshots

N/A


johanneshardt:
Would be great to see this merged! In my case it resulted in an error message, since it tried to run the "find" program, which behaves differently on Windows. With this fix, it instead prompts the user to install fd/ripgrep/something else.

PR #2768, opened 2026-03-14
fix(grep): guard file_sep nil before arithmetic

otakenz:
## Problem

When `rg` outputs a line that does **not** contain a NUL byte separator (e.g. binary-file notices, warning messages, or any non-conforming `rg` output), `item.text:find("\0")` returns `nil`.

The existing nil guard for `file_sep` was placed **after** the lines that perform arithmetic on it, making it unreachable in the error case:

```lua
local file_sep = item.text:find("\0")
local file = item.text:sub(1, file_sep - 1)  -- 💥 crashes here if file_sep is nil
local rest = item.text:sub(file_sep + 1)
item.text = file .. ":" .. rest:gsub(MATCH_SEP, "")
if not file_sep then  -- never reached
  ...
end
```

This produces an unhandled async error:

```
snacks/picker/source/grep.lua:125: attempt to perform arithmetic on local 'file_sep' (a nil value)
```

## Fix

Move the `if not file_sep then` guard to immediately after `find()`, before any arithmetic on `file_sep`:

```lua
local file_sep = item.text:find("\0")
if not file_sep then
  if not item.text:match("WARNING") then
    Snacks.notify.error("invalid grep output:\n" .. item.text)
  end
  return false
end
local file = item.text:sub(1, file_sep - 1)  -- safe
local rest = item.text:sub(file_sep + 1)      -- safe
```

## Trigger conditions

- `rg` encountering binary files (emits `"Binary file <path> matches"` without NUL)
- `rg` warning lines about unreadable paths
- Any `rg` output line that doesn't follow the `filename\0line:col:match` format

### DIFF
--- a/lua/snacks/picker/source/grep.lua
+++ b/lua/snacks/picker/source/grep.lua
@@ -122,23 +122,23 @@ function M.grep(opts, ctx)
         item.cwd = cwd
         -- Split on NUL byte (which comes from rg's -0 flag)
         local file_sep = item.text:find("\0")
-        local file = item.text:sub(1, file_sep - 1)
-        local rest = item.text:sub(file_sep + 1)
-        item.text = file .. ":" .. rest:gsub(MATCH_SEP, "")
         if not file_sep then
           if not item.text:match("WARNING") then
             Snacks.notify.error("invalid grep output:\n" .. item.text)
           end
           return false
         end
+        local file = item.text:sub(1, file_sep - 1)
+        local rest = item.text:sub(file_sep + 1)
         ---@type string?, string?, string?
         local line, col, text = rest:match("^(%d+):(%d+):(.*)$")
         if not (line and col and text) then
           if not item.text:match("WARNING") then
-            Snacks.notify.error("invalid grep output:\n" .. item.text)
+            Snacks.notify.error("invalid grep output:\n" .. item.text:sub(1, file_sep - 1) .. ":" .. item.text:sub(file_sep + 1))
           end
           return false
         end
+        item.text = file .. ":" .. rest:gsub(MATCH_SEP, "")
 
         -- indices of matches
         local from = tonumber(col)


dpetka2001:
This will trigger again #2744. At the minimum `MATCH_SEP` should be removed before sending `item.text` to `vim.notify` as it causes an Error about `Using Blob as a String`.

dpetka2001:
My suggestion would be to revert #2744 and use `local text = item.text:gsub(MATCH_SEP, "")` before the conditional check and send that to `vim.notify` instead or just directly strip `item.text` inside `vim.notify`. It was my bad for causing this, but `MATCH_SEP` should still be stripped from `item.text` before sending it to `vim.notify`.

dpetka2001:
Actually inlining the `gsub` directly in `Snacks.notify` still doesn't work. Only moving `item.text = file .. ":" .. rest:gsub(MATCH_SEP, "")` before the second conditional check works for the issue that #2744 tried to solve. 

This PR will work as it is without causing the problem in #2744. Just tried it out locally.

PS: @otakenz Tnx for catching this and sorry for any inconvenience.

dpetka2001:
It's not the `MATCH_SEP` that causes the error, it's the `file_sep` itself (being a NUL byte). The following works when inlining in the second conditional check
```lua
Snacks.notify.error("invalid grep output:\n" .. item.text:sub(1, file_sep - 1) .. ":" .. item.text:sub(file_sep + 1))
```
The rest can remain as before #2744 was merged.

otakenz:
> It's not the `MATCH_SEP` that causes the error, it's the `file_sep` itself (being a NUL byte). The following works when inlining in the second conditional check
> 
> ```lua
> Snacks.notify.error("invalid grep output:\n" .. item.text:sub(1, file_sep - 1) .. ":" .. item.text:sub(file_sep + 1))
> ```
> 
> The rest can remain as before #2744 was merged.

Thanks, I have updated the commit, does it looks alright now?

folke:
reverted that other commit and removed nul bytes from notify message. ty!

dpetka2001:
@folke It seems that with your latest change https://github.com/LazyVim/LazyVim/issues/7025 is still present. I get the error described there. The notification with `invalid grep output` does not trigger. 

Would you mind using `Snacks.notify.error("invalid grep output:\n" .. item.text:sub(1, file_sep - 1) .. ":" .. item.text:sub(file_sep + 1))` instead? With that change the proper notification is shown instead of the Neovim error.

PS: Or move line 141 just before the second conditional check on line 135. I don't think that would cause any error, right?

dpetka2001:
@folke I also tried `gsub("%z", "")` and it also worked if you'd like to keep it inline. For some reason the `\0` representation fails (I'm testing at the time on a Windows machine, don't know if that's maybe a reason).

folke:
@dpetka2001 just chnaged it to `%z`, thanks!

PR #2769, opened 2026-03-15
fix(rename): use colon syntax for LSP client methods and update get_clients

lgick:
## Description

This PR fixes deprecation warnings appearing in Neovim nightly (0.11+) when renaming files. 

Changes:
- Replaced `client.supports_method`, `client.request_sync`, and `client.notify` with the method call syntax `client:method(...)`.
- Replaced the deprecated `vim.lsp.get_active_clients()` with `vim.lsp.get_clients()`.

These changes align with the latest Neovim LSP API requirements.

## Related Issue(s)

Fixes deprecation warnings reported by `:checkhealth`.

## Screenshots
N/A (LSP API update)

### DIFF
--- a/lua/snacks/picker/source/lsp/init.lua
+++ b/lua/snacks/picker/source/lsp/init.lua
@@ -30,39 +30,6 @@ function M.symbol_kind(kind)
   return kinds[kind] or "Unknown"
 end
 
---- Neovim 0.11 uses a lua class for clients, while older versions use a table.
---- Wraps older style clients to be compatible with the new style.
----@param client vim.lsp.Client
----@return vim.lsp.Client
-local function wrap(client)
-  local meta = getmetatable(client)
-  if meta and meta.request then
-    return client
-  end
-  ---@diagnostic disable-next-line: undefined-field
-  if client.wrapped then
-    return client
-  end
-  local methods = { "request", "supports_method", "cancel_request", "notify" }
-  -- old style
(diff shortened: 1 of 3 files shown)


dpetka2001:
This is a breaking change and will drop support for Neovim versions <0.11. No idea until which version maintainer wants to support. Current supported version is 0.9.4 according to README.

But an alternative would be to export https://github.com/folke/snacks.nvim/blob/a049339328e2599ad6e85a69fa034ac501e921b2/lua/snacks/picker/source/lsp/init.lua#L37-L64 into `snacks/util/init.lua` and then use that. Also I do not get any deprecation warnings about `vim.lsp.get_active_clients()` in `checkhealth`. Only inline diagnostics in the file itself, but that should not be really a problem since it depends on the Neovim version you have installed.

lgick:
Done! I exported the existing wrap function from picker/source/lsp/init.lua into snacks/util/init.lua as you suggested, and updated both picker and rename to use Snacks.util.wrap. Also reverted the get_clients fallback for Neovim 0.9 compatibility.

dpetka2001:
Please do note i'm just a simple user and under no circumstances do i make decisions in this project. My previous comment was more like a suggestion/note. Maybe the maintainer would be in favor of dropping support for Neovim <0.11. I've literally no idea. 


He's the one with final say and make final decision.

lgick:
Thanks for the suggestion, it was helpful 🤝

Issue #2332, opened 2025-10-23
bug: Terminal probe sequences are inserted as literal text when tmux with extended-keys on is used.


twio142:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.4

### Operating system/version

macOS 15.7.1

### Describe the bug

When running Neovim inside `tmux` with `extended-keys on` (which is necessary for combos like `<C-CR>` and `<S-CR>` to work), the escape sequences that `snacks.nvim` sends to probe for Kitty image protocol are not handled silently. Instead, they are inserted directly into the current buffer as literal keys.

This behavior appears to have been introduced in commit 43261baf.

This seems to be the direct cause:
https://github.com/folke/snacks.nvim/blob/277ba028b3b764ee49c9be8fd4d1282aff86812a/lua/snacks/image/terminal.lua#L228-L233

### Steps To Reproduce

1. Launch `tmux` with `set -g extended-keys on`
2. Run `nvim -u repro.lua repro.lua` with the `repro.lua` below.
3. Yank any text
4. Run `:CopilotChatToggle` (which opens a new buffer and enters insert mode)
5. It either auto-pastes the previously yanked text, or enters `ostty 1.2.2` (I'm on ghostty 1.2.2)

P.S. I also tested the minimal setup with `avante.nvim`, which also has a command opening a new buffer and entering insert mode, and the same behavior occurs. I didn't want to involve another plugin, but it seems to be the most consistent way of reproduction. Both plugins share the same dependency `plenary.nvim`.

### Expected Behavior

Not this.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    { "folke/snacks.nvim", opts = { image = { enabled = true } } },
    {
      'CopilotC-Nvim/CopilotChat.nvim',
      cmd = { 'CopilotChatToggle' },
      dependencies = { 'nvim-lua/plenary.nvim' },
    },
  },
})
```

folke:
Fixed in a0d75be.

[...]

PR #2790, opened 2026-03-31
fix(image): handle tmux extended-keys always

dvic:
## Description
Treat `tmux show -g extended-keys` returning `extended-keys always` the same as `extended-keys on` in image terminal detection.

Without this, snacks misses the tmux `client_termname` workaround and falls back to the terminal query / `TermResponse` path.

Verified locally with tmux 3.6a + Ghostty, where this fixed markdown picker freezes caused by image terminal detection.

## Related Issue(s)
Related to #2332

## Screenshots
N/A


vvatikiotis:
I can confirm that the fix works. I've patched `terminal.lua` with the fix, prior to seeing this PR and it works as expected.

Jimfarrugia:
> I can confirm that the fix works. I've patched `terminal.lua` with the fix, prior to seeing this PR and it works as expected.

I'll second this.  I'm using kitty terminal and tmux with `extended-keys always` set.

dkarter:
@folke this seems very reasonable 🙏🏼 would you accept this PR?

PR #2799, opened 2026-04-03
fix(picker): change order of `cmd` for fd/fdfind and adjust health files

dpetka2001:
## Description
See https://github.com/LazyVim/LazyVim/issues/7093

This has come up multiple times now. I believe it's a small QOL for new users who don't know how to interpret things. Because in the current state, if a user has fdclone installed checkhealth will fail with `attempt to compare nil with a table`. If both fdclone and fd-find are installed, a user has to symlink fdfind to fd.

Feel free to disregard if deemed out of scope. Totally understandable.

## Related Issue(s)
https://github.com/LazyVim/LazyVim/issues/7093

### DIFF
--- a/lua/snacks/health.lua
+++ b/lua/snacks/health.lua
@@ -93,7 +93,24 @@ function M.have_tool(tools)
         if vim.fn.executable(cmd) == 1 then
           local version = tool.version == false and "" or vim.fn.system(cmd .. " --version") or ""
           version = vim.trim(vim.split(version, "\n")[1])
-          if tool_version and tool_version > vim.version.parse(version) then
+          local parsed_version = nil
+          if tool.version ~= false then
+            parsed_version = vim.version.parse(version)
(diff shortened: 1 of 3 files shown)


lua/snacks/health.lua
```diff
@@ -103,6 +120,9 @@ function M.have_tool(tools)
             version_ok = true
           end
           found = true
+          if vim.deep_equal(cmds, { "fdfind", "fd" }) then
+            break
+          end
```
dpetka2001, line 125: This is needed in case users have both `fdclone` and `fd-find` installed. The binary `fd` (from fdclone) won't pass the checkhealth and will abrubtly cause the rest of checkhealth to be cancelled. 

So, in case `fdfind` passes checkhealth, I believe it's safe to assume we don't need to also checkhealth for `fd` (which will most probably be the binary from fdclone).

PR #2802, opened 2026-04-04
fix(image): check cursor is on the right line in at_cursor

antinomie8:
## Description

I noticed the floating images sometimes appeared even when my cursor was not on the image's line (see screenshot below). This is a trivial fix to ensure the cursor is on the same line as a single-line image treesitter node.

## Screenshots

Currently, the image is shown when my cursor is above it:
image

### DIFF
--- a/lua/snacks/image/doc.lua
+++ b/lua/snacks/image/doc.lua
@@ -355,7 +355,7 @@ function M.at_cursor(cb)
       local range = img.range
       if range then
         if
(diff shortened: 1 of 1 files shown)


anuragkumar2921:
I'm also facing the same issue and this change fix the issue. Requesting maintainer to please merge the change. Thanks   @antinomie8 for the fix. cheers !!

PR #2831, opened 2026-04-17
feat(toggle): added codelens toggle

cub-uanic:
## Description

Adds a `codelens` toggle.

## Related Issue(s)

Required for LazyVim/LazyVim#7113

## Screenshots

None.

### DIFF
--- a/lua/snacks/toggle.lua
+++ b/lua/snacks/toggle.lua
@@ -352,6 +352,20 @@ function M.scroll()
   })
 end
 
+function M.codelens()
+  return M.new({
+    id = "codelens",
+    name = "Codelens",
+    has = "codeLens",
+    get = function()
+      return vim.lsp.codelens.is_enabled()
+    end,
+    set = function(state)
+      vim.lsp.codelens.enable(state)
+    end,
+  })
+end
+
 function M.zen()
   return M.new({
     id = "zen",


hellupline:
I liked the idea and decided to test on my current setup, but `vim.lsp.codelens.is_enabled()` was returning false until I toggled twice

I did this small update to fix this condition

```lua
Snacks.toggle
  .new({
    id = 'codelens',
    name = 'Codelens',
    has = 'codeLens',
    get = function() return vim.lsp.codelens.is_enabled({ bufnr = vim.api.nvim_get_current_buf() }) end,
    set = function(state) vim.lsp.codelens.enable(state, { bufnr = vim.api.nvim_get_current_buf() }) end,
  })
  :map('<leader>uH')
```

PFiS1737:
Here are some suggestions from my side:

1. As far as I know, there is no parameter field called `has`.

2. We can follow the design of `M.inlay_hints`, whose interface is almost identical, and add `{ bufnr = 0 }`. This should also resolve the issue mentioned above.

https://github.com/folke/snacks.nvim/blob/ad9ede6a9cddf16cedbd31b8932d6dcdee9b716e/lua/snacks/toggle.lua#L206-L217

3. It would be better to place `M.codelens` near `M.inlay_hints`, since both are handling LSP-related functionality.

4. Optionally, consider providing backward compatibility. The README states that this plugin targets Neovim >= 0.9.4, while the API used in this PR was introduced in 0.12. However, note that `M.inlay_hints` relies on an API introduced in 0.10 and does not perform any compatibility checks either. (If you’ve looked at LazyVim’s implementation, you’ll see that such checks are done at usage time.)

https://github.com/LazyVim/LazyVim/blob/7c1301b89563198d1b2a295e673f6a7205adc076/lua/lazyvim/config/keymaps.lua#L160-L162

PR #2864, opened 2026-05-21
fix(image): apply tmux extended-keys workaround for `always` too

yakovlievv:
## Description

`tmux show -g extended-keys` returns `on`, `off`, or `always`. The existing workaround for terminal-response leaks only triggered when the value was `on`, so users with `set -s extended-keys always` still saw the `\033[>q` XTVERSION reply leak into the active window as literal text (e.g. `tty(0.46.2)`), often also tripping bracketed paste.

Match `always` in addition to `on` so the workaround engages for both.

## Related Issue(s)

Refs: #2332

## Screenshots

 N/A — terminal escape leak, no visual change.


vvatikiotis:
same as https://github.com/folke/snacks.nvim/pull/2790

PR #2873, opened 2026-05-29
fix(picker,explorer): resolve bare executable names on Windows before uv.spawn

### LINKED ISSUE #2872: bug(picker): uv.spawn fails to resolve rg/fd on Windows despite healthcheck passing

### Neovim version (nvim -v)

0.12.2

### Operating system/version

Windows 11

### Describe the bug

On native Windows Neovim, Snacks.picker.files() / Snacks.picker.grep() can fail with `Failed to spawn fd` or `Failed to spawn rg` even though `:checkhealth` snacks reports both rg and fd as available. 

The failing call appears to be `uv.spawn()` in lua/snacks/picker/source/proc.lua, and this happens even though `:lua print(vim.fn.exepath("rg"))` works, where it resolves to a scoop shim: `C:\Users\user\scoop\shims\rg.EXE`

The issue seems to be that snacks.picker.source.proc passes the bare command name, e.g. "rg" or "fd", directly to `uv.spawn()`. On this Windows setup, raw `uv.spawn("rg", ...)` returns ENOENT, while resolving the executable first through `vim.fn.exepath()` fixes the picker.

### Steps To Reproduce

1. Install `ripgrep` and `fd` through scoop: ```scoop install ripgrep fd```
2. Confirm they work in powershell: `rg --version`, `fd --version`
3. Confirm neovim can resolve them: 
```
:lua print(vim.fn.exepath("rg"))
:lua print(vim.fn.exepath("fd"))
```
4. Confirm `:checkhealth snacks` can find `fd` and `rg`
5. Run `:lua Snacks.picker.files()` or `:lua Snacks.picker.grep()` 
6. Picker fails with `Failed to spawn fd` / `Failed to spawn rg`.

### Expected Behavior

If `vim.fn.exepath("rg")` / `vim.fn.exepath("fd")` can resolve the executable, and healthcheck reports the tools as available, then picker sources should be able to spawn them.

On Windows, `snacks.picker.source.proc` should probably resolve bare command names before calling `uv.spawn()`. For example, I was able to fix the issue with something like
```lua
local function resolve_spawn_cmd(cmd)
  if vim.loop.os_uname().sysname ~= "Windows_NT" then
    return cmd
  end

  if type(cmd) ~= "string" then
    return cmd
  end

  if cmd:find("[/\\]") then
    return cmd
  end

  local resolved = vim.fn.exepath(cmd .. ".exe")
  if resolved ~= "" then
    return resolved
  end

  resolved = vim.fn.exepath(cmd)
  if resolved ~= "" then
    return resolved
  end

  return cmd
end
```
Then changing the `uv.spawn` line to
```lua
uv.spawn(resolve_spawn_cmd(opts.cmd), spawn_opts, ...)
```

This probably affects anything that calls `uv.spawn()` with a bare command name on Windows, not just rg/fd.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
  },
})
```

truffle-dev:
## Description

On native Windows Neovim, `Snacks.picker.files()` / `Snacks.picker.grep()` can fail with `Failed to spawn rg` / `Failed to spawn fd` even when `:checkhealth snacks` reports both as available and `:lua print(vim.fn.exepath("rg"))` resolves to a working shim (e.g. `C:\Users\user\scoop\shims\rg.EXE`). Same shape applies to `Snacks.explorer`'s git status poller calling bare `git`, and to anything else that goes through `Snacks.spawn`.

The root cause is that libuv's `spawn` does not search `PATH` for bare command names on Windows the way it does on POSIX. The reporter (#2872) traced the failing call to `uv.spawn(opts.cmd, ...)` in `lua/snacks/picker/source/proc.lua`, confirmed the picker recovers when the executable is resolved through `vim.fn.exepath()` first, and noted that the same pattern probably affects every `uv.spawn` call site.

There are three of those today:

- `lua/snacks/util/spawn.lua` — the `Proc:run` helper, used by `Snacks.image.convert`, `Snacks.gh.api`, and anything else that calls `Snacks.spawn`
- `lua/snacks/picker/source/proc.lua` — `picker.files`, `picker.grep`, and every other process-backed source
- `lua/snacks/explorer/git.lua` — the git status poller behind the explorer's status column

This PR adds a small `Snacks.spawn.resolve_cmd(cmd)` helper and wraps the command at each of those sites. The helper:

- Returns the input unchanged on non-Windows systems (libuv's POSIX path search already works there)
- Returns the input unchanged if the value already contains `/` or `\` (it's a path, no resolution needed)
- On Windows, tries `vim.fn.exepath(cmd .. ".exe")` first, then `vim.fn.exepath(cmd)`, returning the input unchanged if neither resolves

No behavior change on Linux/macOS. On Windows it does exactly what the issue's snippet did, just in one shared helper instead of inline at each site.

## Testing

- `stylua --check` clean on the three modified files plus the new spec
- Luac syntax check clean
- New `tests/spawn_spec.lua` covers the platform guard, defensive inputs (nil, empty, POSIX path, Windows-style backslash path), and the no-op on non-Windows for bare names. The Windows-specific resolution path is exercised by the reporter's scoop setup in #2872; I don't have a Windows CI runner here, so the new spec only locks the contract that exists on the platforms CI does run on
- No new dependencies; only `vim.fn.exepath` and `uv.os_uname` are used, both available in every supported Neovim

Closes #2872

### DIFF
--- a/lua/snacks/explorer/git.lua
+++ b/lua/snacks/explorer/git.lua
@@ -1,4 +1,6 @@
 ---@diagnostic disable: missing-fields
+local Spawn = require("snacks.util.spawn")
+
 local M = {}
 
 ---@class snacks.explorer.git.Status
@@ -55,7 +57,7 @@ function M.update(cwd, opts)
   local output = ""
   local stdout = assert(uv.new_pipe())
   local handle ---@type uv.uv_process_t
-  handle = uv.spawn("git", {
+  handle = uv.spawn(Spawn.resolve_cmd("git"), {
     stdio = { nil, stdout, nil },
     cwd = root,
     hide = true,
--- a/lua/snacks/picker/source/proc.lua
+++ b/lua/snacks/picker/source/proc.lua
@@ -1,5 +1,6 @@
 ---@diagnostic disable: await-in-sync
 local Async = require("snacks.picker.util.async")
+local Spawn = require("snacks.util.spawn")
 
 local M = {}
 
@@ -58,7 +59,7 @@ function M.proc(opts, ctx)
 
     local handle ---@type uv.uv_process_t
     ---@diagnostic disable-next-line: missing-fields
-    handle = uv.spawn(opts.cmd, spawn_opts, function(code, _signal)
+    handle = uv.spawn(Spawn.resolve_cmd(opts.cmd), spawn_opts, function(code, _signal)
       if not aborted and code ~= 0 and opts.notify ~= false then
         local full = { opts.cmd or "" }
         vim.list_extend(full, opts.args or {})
--- a/lua/snacks/util/spawn.lua
+++ b/lua/snacks/util/spawn.lua
@@ -5,6 +5,34 @@ local M = {}
 
 local uv = vim.uv or vim.loop
 
+local is_windows = (uv.os_uname().sysname or ""):lower():find("windows") ~= nil
+
+--- Resolve a bare executable name to an absolute path on Windows.
+--- libuv's `spawn` does not search `PATH` for bare names on Windows the way
+--- it does on POSIX, so a command like `rg` may return ENOENT even when
+--- `vim.fn.exepath("rg")` resolves it (e.g. through a scoop shim).
+--- Returns the input unchanged on non-Windows systems, when given a path
+--- (anything containing `/` or `\`), or when resolution fails.
+---@param cmd string
+---@return string
+function M.resolve_cmd(cmd)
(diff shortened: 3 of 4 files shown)


truffle-dev:
Still relevant: #2872 is open and there's no PATH resolution on main, so bare `rg`/`fd`/`git` still fail to spawn on native Windows even when exepath resolves them. This is rebased, mergeable, three checks green. If three call sites is too broad, happy to narrow it to one helper in `spawn.lua` and route the picker/explorer through that instead. Ready whenever you have a slot.

PR #2874, opened 2026-05-29
fix(picker.projects): pass --no-ignore-vcs so fd 9.0.0 surfaces .git

### LINKED ISSUE #2814: bug: Snacks.picker.projects() does not correctly locate git projects with `fd` v9.0.0

### Neovim version (nvim -v)

NVIM v0.12.0-dev-2719+g9c5d8b475e

### Operating system/version

Ubuntu 24.04.4 LTS

### Describe the bug

`.git` projects are not correctly discovered using the project picker as a result of the quoted breaking change in `fd` v9.0.0 (officially distributed in Ubuntu 24.04.4 LTS). 

I realise that this has already been reverted in `fd` versions >= v10.0.0 but raising this incase it either warrants either extra documentation, or instead adding the `--no-ignore` flag to the list of used arguments.

Relevant changelog (https://github.com/sharkdp/fd/releases/tag/v9.0.0)
> Breaking: .git/ is now ignored by default when using --hidden / -H, use --no-ignore / -I or --no-ignore-vcs to override, see https://github.com/sharkdp/fd/issues/1387 and https://github.com/sharkdp/fd/pull/1396 (@skoriop)

### Steps To Reproduce

- install `fd` v9.0.0
- create git project in `~/projects/foobar`
- run `:lua Snack.picker.projects()`

### Expected Behavior

Picker should discover the `foobar` project.

### Repro

```lua

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

Booyaka101:
## Description

fd 9.0.0 ([sharkdp/fd#1396](https://github.com/sharkdp/fd/pull/1396)) made `.git/` ignored by default even with `--hidden`, breaking `Snacks.picker.projects()` discovery on every system where that build is installed. Reverted in fd 10.0.0 but Ubuntu 24.04 LTS still ships fd 9.0.0 in its archive, so users on that distro can't find any `.git` projects.

The fd changelog calls out `--no-ignore-vcs` as the targeted override. Adding it to the args list in `lua/snacks/picker/source/recent.lua:M.projects` is a no-op on fd 8.x and fd 10+ for the existing `-H -g {.git,...}` use.

### Verification

`cargo install --version 9.0.0 fd-find` on Windows, two empty git projects under `/tmp/fd-test`:

| | output |
| --- | --- |
| pre-fix (`fd -H -t f -t s -t d --max-depth 2 --follow --absolute-path -g "{.git,_darcs,.hg,.bzr,.svn,package.json,Makefile}" /tmp/fd-test`) | (empty) — reproduces |
| post-fix (`+ --no-ignore-vcs`) | `proj1/.git/` and `proj2/.git/` — fixed |

Side-effect check on a `node_modules` + `.gitignore` project:

| | max_depth=2 (default) | max_depth=3 |
| --- | --- | --- |
| pre-fix | top-level `package.json` only | same |
| post-fix | top-level `package.json` + `.git/` | + the gitignored `node_modules/lodash/package.json` |

No regression at the default depth. At deeper custom depths the flag's documented behavior (descending into gitignored dirs) becomes observable, which is what someone passing `max_depth=4+` for a project discovery use is reasonably asking for.

## Related Issue(s)

- Fixes #2814

## Screenshots

N/A

### DIFF
--- a/lua/snacks/picker/source/recent.lua
+++ b/lua/snacks/picker/source/recent.lua
@@ -63,6 +63,7 @@ M.recent = M.files
 function M.projects(opts, ctx)
   local args = {
     "-H",
+    "--no-ignore-vcs", -- fd 9.0.0 hides .git/ with -H, this restores it (sharkdp/fd#1396)
     "-t",
     "f",
     "-t",


Booyaka101:
Closing as a duplicate of #2815 by @thom-jenks (the original reporter), which has been open since April with the same scope. Apologies for the noise — I should have searched existing PRs before opening this one.

For what it's worth, the only technical difference is the flag choice: `--no-ignore-vcs` is the targeted override the fd changelog recommends and leaves `.fdignore` / global ignore behavior intact, where `--no-ignore` disables all ignore files. Either fixes #2814; I'll leave that call to whoever reviews #2815.

Issue #2539, opened 2025-11-20
bug(picker): using `vim.ui.select` with snacks picker, get Invalid 'height': Number is not integral


milanglacier:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

11.5

### Operating system/version

MacOS 26

### Describe the bug

The following example code will give the following error:

```lua
Example_Choices = {}
for i = 1, 1000 do
    Example_Choices[i] = i
end

Example_UI_select = function()
    vim.ui.select(Example_Choices, {}, function() end)
end
```

```
E5108: Error executing lua ...ar/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:810: Invalid 'height': Number is not integral                            
stack traceback:                                                                                                                                                
        [C]: in function 'nvim_win_set_config'                                                                                                                  
        ...ar/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:810: in function 'update'                                                                   
        ....local/share/nvim/lazy/snacks.nvim/lua/snacks/layout.lua:288: in function 'update'                                                                   
        ....local/share/nvim/lazy/snacks.nvim/lua/snacks/layout.lua:583: in function 'show'                                                                     
        .../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:489: in function 'show'                                                                     
        .../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:744: in function 'update'                                                                   
        ...nvim/lazy/snacks.nvim/lua/snacks/picker/core/matcher.lua:141: in function 'run'                                                                      
        .../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:845: in function 'find'                                                                     
        .../nvim/lazy/snacks.nvim/lua/snacks/picker/core/picker.lua:142: in function 'select'                                                                   
        /Users/northyear/Desktop/dotfiles/config/nvim/init.lua:26: in function 'fff'                                                                            
        [string ":lua"]:1: in main chunk
```

Example Snacks config:

```lua
opts=  {
        layout = {
            preset = 'vertical',
            layout = {
                height = 0.85,
            },
        },

}

```

[...]

PR #2871, opened 2026-05-29
fix(picker.select): floor list height to avoid fractional dim

Booyaka101:
## Description

`vim.o.lines * 0.8 - 10` in `lua/snacks/picker/select.lua:46` is fractional whenever `vim.o.lines` is not a multiple of 5 (e.g. 37 -> 19.6). When `#items` exceeds that bound, the surrounding `math.min` returns the fractional value and it lands on the `list` box height; `M:dim()`'s `size()` helper only floors when `s == 0` or `s < 1`, so the fractional flows through `nvim_win_set_config`:

```
E5108: Error executing lua .../snacks.nvim/lua/snacks/win.lua:810: Invalid 'height': Number is not integral
stack traceback:
        [C]: in function 'nvim_win_set_config'
        .../snacks.nvim/lua/snacks/win.lua:810: in function 'update'
        .../snacks.nvim/lua/snacks/layout.lua:288: in function 'update'
        .../snacks.nvim/lua/snacks/layout.lua:583: in function 'show'
        .../snacks.nvim/lua/snacks/picker/core/picker.lua:489: in function 'show'
        .../snacks.nvim/lua/snacks/picker/core/picker.lua:744: in function 'update'
```

Floor the bound at the source.

### Verification

`nvim --headless -u repro.lua --cmd "set lines=37 columns=120"` on Neovim 0.12.2 / Windows. Tap on `nvim_win_set_config` logs any non-integer dim before the call reaches Neovim:

| variant                                     | items | picker height | pre-fix              | post-fix |
| ------------------------------------------- | ----- | ------------- | -------------------- | -------- |
| default `select` preset                     | 50    | n/a           | ok                   | ok       |
| user 0.85 (the issue's repro)               | 100   | 0.85          | FAIL `height=21.6`   | ok       |
| user 0.6                                    | 5     | 0.6           | ok                   | ok       |
| user 0.95                                   | 1000  | 0.95          | FAIL `height=21.6`   | ok       |
| user 0.5, empty list                        | 0     | 0.5           | ok (no results)      | ok (no results) |

Pre-fix: 2 of 5 variants reproduce. Post-fix: 0 non-integer dim calls across all 5, no regression on the previously-passing ones. The `height=21.6` matches the issue's failure (`vim.o.lines=37, cmdheight=1 -> inner area 36, 36 * 0.6 = 21.6` after border accounting).

## Related Issue(s)

- Fixes #2539

## Screenshots

N/A

PR #2916, opened 2026-07-20
fix(picker): floor list height to prevent non-integral height error in vim.ui.select

zeroDivisible:
## Description

`Snacks.picker.select` computes the list height as `vim.o.lines * 0.8 - 10` without flooring. When `vim.o.lines` isn't a multiple of 5, the result is fractional and gets passed straight to `nvim_win_set_config`, which requires an integral height for floating windows. Any `vim.ui.select` call with enough items to hit the cap triggered E5108, blocking the picker from rendering.

This commit wraps the cap in `math.floor` to guarantee an integer.

## Related Issue(s)
I hit this bug in my config but when filing an issue, I found a preexisting one: https://github.com/folke/snacks.nvim/issues/2539.

This fix should solve it.

## Screenshots
N/A

### DIFF
--- a/lua/snacks/picker/select.lua
+++ b/lua/snacks/picker/select.lua
@@ -43,7 +43,7 @@ function M.select(items, opts, on_choice)
         -- Fit list height to number of items, up to 10
         for _, box in ipairs(layout.layout) do
           if box.win == "list" and not box.height then
-            box.height = math.max(math.min(#items, vim.o.lines * 0.8 - 10), 2)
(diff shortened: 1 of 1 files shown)


kluen:
I believe the change is identical to this 2 months old PR: https://github.com/folke/snacks.nvim/pull/2871

zeroDivisible:
It is @kluen. I'm going to close this and hope the other one gets merged.

Issue #1417, opened 2025-02-25
bug: keymaps not applied correctly in preview window when open buffer is focused in smart picker 


n-crespo:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.11.0-dev

### Operating system/version

Ubuntu 24.04.2 WSL

### Describe the bug

When using the "smart" picker, the 'focus_input' action doesn't work when the cursor is in a preview that is of a buffer that is currently open. For all other preview windows, this works as expected.  

### Steps To Reproduce

1. map something, e.g. <tab> to focus_input in snacks config (in input window section set the action to "focus_preview" and in preview section "focus_input")
2.  Open any file e.g. test.lua
3. Open smart picker
4. Select the file test.lua (has some green number next to it)
5. hit <tab> to focus the preview window
6. hit <tab> again to go back to the input (nothing happens)
7. :q to close the picker
8. Error thrown (below) (could not reproduce with repro.lua)
```
   Error  10:28:43 PM msg_show.lua_error Error detected while processing BufEnter Autocommands for "<buffer=9>":
   Error  10:28:43 PM msg_show.emsg Error executing lua callback: ...as/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:468: ...e/nvim/lazy/snacks.nvim/lua/snacks/picker/core/input.lua:52: Invalid 'name': Expected Lua string
stack traceback:
	[C]: in function 'error'
	...as/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:468: in function <...as/.local/share/nvim/lazy/snacks.nvim/lua/snacks/win.lua:464>
```

### Expected Behavior

When <tab> is pressed in the preview window, focus is returned to the input. 

### Repro

```lua
---@diagnostic disable: missing-fields
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      "folke/snacks.nvim",
      opts = {
        -- set the dir to the directory of the current buffer if possible, then
        -- the dir of the alternate buffer, then the cwd
        picker = {
          formatters = {
[...]

Issue #773, opened 2025-01-28
bug: Picker buffers not always sorting correctly on most recently used


gwww:
### Did you check docs and existing issues?

- [x] I have read all the snacks.nvim docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched the existing issues of snacks.nvim
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.3

### Operating system/version

MacOS 15.2

### Describe the bug

When using the picker to select a buffer the sort order is not as expected. Details below.

### Steps To Reproduce

I am using the repro script below.

I have a directory for reproducing this issue. It has 4 empty files, named aaa, bbb, ccc, and ddd.

Start neovim with `nvim -u repro.lua aaa bbb ccc ddd`. `aaa` is the current buffer being edited.

Invoke the picker using `<space>fb`. The following is displayed:
```
bbb
ccc
ddd
```
Select `ddd` (I use the arrow keys to move the selection). `ddd` is the current buffer being edited, as expected.

Invoke the picker using `<space>fb`. The following is displayed:
```
aaa
bbb
ccc
```
Select `ccc`. `ccc` is the current buffer being edited, as expected.

Invoke the picker using `<space>fb`. The following is being displayed:
```
aaa
ddd
bbb
```
I expect the order to be:
```
ddd
aaa
bbb
```
I would expect `ddd` to be at the top but `aaa` is at the top. `aaa` was using two buffers ago.

### Expected Behavior

See above.

Note that telescope does not exhibit this behaviour.

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system "curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua")()

require("lazy.minit").repro {
  spec = {
    {
      "folke/snacks.nvim",
      opts = {
        picker = {},
      },
      keys = {
        {
          "<space>fb",
          function() require("snacks").picker.buffers { current = false, sort_lastused = true } end,
          desc = "Buffers",
        },
      },
    },
  },
}
```

folke:
The reason is that previewing the buffer also triggers a `lastused` update.

Snacks uses the existing buffer when displaying files if available so that you see the preview exactly how it should be,
and only resorts to a scratch buffer when needed.

Telescope always uses scratch buffers.

[...]

PR #2927, opened 2026-08-09
feat(picker.preview): add option to opt out of real-buffer preview

fmorroni:
## Description

Picker previews can be of two types: scratch buffers or real buffers. Scratch buffers are used when the file isn't already open, while real buffers are used when the file is already loaded.

Using a real buffer means the preview inherits the buffer's existing state and keymaps, which can interfere with picker-specific keymaps (see https://github.com/folke/snacks.nvim/issues/1417). It also updates the buffer's last-used timestamp, which can affect picker sorting (see https://github.com/folke/snacks.nvim/issues/773).

For my use cases, having the real-buffer preview has no benefits. Other users also seem to feel the same way (https://github.com/folke/snacks.nvim/issues/773#issuecomment-4467371981).

This PR adds a `use_real_buffer` option to allow users to opt out of real-buffer previews and always use scratch buffers instead. You would opt out by setting `picker.previewers.file.use_real_buffer = false`.

## Related Issue(s)

https://github.com/folke/snacks.nvim/issues/1417
https://github.com/folke/snacks.nvim/issues/773

### DIFF
--- a/lua/snacks/picker/config/defaults.lua
+++ b/lua/snacks/picker/config/defaults.lua
@@ -198,6 +198,7 @@ local defaults = {
       max_size = 1024 * 1024, -- 1MB
       max_line_length = 500, -- max line length
       ft = nil, ---@type string? filetype for highlighting. Use `nil` for auto detect
+      use_real_buffer = true ---@type boolean use actual buffer as preview when the file is already loaded (default: true)
     },
     man_pager = nil, ---@type string? MANPAGER env to use for `man` preview
   },
--- a/lua/snacks/picker/preview.lua
+++ b/lua/snacks/picker/preview.lua
@@ -101,7 +101,7 @@ function M.file(ctx)
     vim.fn.bufload(ctx.item.buf)
   end
 
-  if ctx.item.buf and vim.api.nvim_buf_is_loaded(ctx.item.buf) then
+  if ctx.item.buf and vim.api.nvim_buf_is_loaded(ctx.item.buf) and ctx.picker.opts.previewers.file.use_real_buffer then
     if not title then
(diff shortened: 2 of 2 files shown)


dpetka2001:
You might also want to make the following addition
```diff
diff --git a/lua/snacks/picker/source/buffers.lua b/lua/snacks/picker/source/buffers.lua
index f7a7ee7a..47e9f349 100644
--- a/lua/snacks/picker/source/buffers.lua
+++ b/lua/snacks/picker/source/buffers.lua
@@ -47,6 +47,9 @@ function M.buffers(opts, ctx)
   end
   if opts.sort_lastused then
     table.sort(items, function(a, b)
+      if a.info.lastused == b.info.lastused then
+        return a.buf < b.buf
+      end
       return a.info.lastused > b.info.lastused
     end)
   end
```
because otherwise the order will still be non-deterministic for same `lastused` timestamps. For example I opened Neovim with multiple buffers, but only the first one was loaded. The rest had `lastused = 0` and `table.sort` does not return the same order if you delete a buffer for example. And after you delete a buffer the picker refreshes and `table.sort` sorts items again but possibly not with same order always if all have same `lastused` value. We need a tiebreaker.

fmorroni:
Hmm yeah that makes sense. Although I think that change should be added regardless of whether this PR is accepted or not, so maybe it should go in its own PR?

That being said, I have no issues with adding it here if that seems best.

dpetka2001:
I'm just a simple user myself like you. I just made the suggestion because you mentioned in your OP about how `lastused` affects picker sort order. Without this change the sort order will still be affected. 

Whether you want to make a different PR or add it here is up to you. Not making any decisions myself.

